PR #21, opened 2021-02-02
Implement VTT DAG process, resolves #20

### LINKED ISSUE #20: Add MongoDB identifier for device type to separate DAG tasks
The DAG's currently retrieve all tasks with non-downloaded data, regardless of device. To distinguish this per device - due to their specialists tasks - we need another identifier to query the database for tasks.

In `dreem_dag()`, it would then be:

```records_not_downloaded("DRM")```

And in db > main.py:

```python
def records_not_downloaded(type: str) -> [Record]:
     docs = _db.records.find({"is_downloaded": False, "device": type})
     return [Record(**doc) for doc in docs]
```

As we are already using acronyms for file types (e.g. DRM for Dreem), this type can follow the same annotation (i.e. DRM, SMA, etc.)

Possibly this issue will be resolved when merging the `vtt` branch

davidverweij:
This PR implements the VTT SMA data download, processing and upload pipeline following the Dreem implementation in #14. It hereby uses the UCAM DB mock service (#18) to translate VTT hashes into Patient ID's and more. The library `boto3` is used to connect to the AWS S3 bucket, and `mypy_boto3_s3` for typing. 

Note that the folder structure in S3 buckets are symbolic. Instead of navigating the bucket, connecting [returns a list of ObjectSummary](https://boto3.amazonaws.com/v1/documentation/api/1.9.42/reference/services/s3.html?highlight=object#S3.Bucket.objects) with keys, where keys represent that symbolic structure, i.e.:

```python
[
    ObjectSummary(key="date/raw/patienthash/file.zip", ...), 
    ObjectSummary(key="date/raw/patienthash/file.nfo", ...),
....
]
```
Note that the above two files are in the same 'folder', but result in two almost identical keys. The logic in `lib > vttsma.py` filters these duplicates to determine the present patients and work from there.

## Other changes/notes
- added a method in `dmpy.py` to .zip and immediately remove the original folder. _This is used to combined the downloaded files from the S3 bucket in one file. `task_prepare_data` does not cope with directories._
- the S3 bucket has various data dump folders. Currently the logic does not look across dumps to determine if patients exists in both. It does collect the various data dumps associated for each patient - but when the `Record` is created for the MongoDB, only one dump date is stored. _See `vttsma_dump_date=item['dumps'][0],`_
- A list of `DeviceTypes` is added to `utils`. Enums were used for code readability (and providing code suggestions in your editor) and to avoid hardcoding mistakes - [and other benefits](https://www.pybootcamp.com/blog/how-to-use-enum-in-python/). **I did change TMA to SMA from [the original list](https://github.com/ideafast/ideafast-patientapp/blob/master/src/styles/colors.js#L14-L26) - do correct me if preferred otherwise.**

## Test
- Add the latest `ucam_db.csv` to your root folder
- update `.dtransfer.env` to include the keys below, where the VTTSMA global ID is the ID for all SMA 'devices' for the DMP as communicated on email.
   - VTTSMA_AWS_ACCESSKEY=""
   - VTTSMA_AWS_SECRET_ACCESSKEY=""
   - VTTSMA_AWS_BUCKET_NAME=""
   - VTTSMA_GLOBAL_DEVICE_ID=""
 - run `poetry install` to include `boto3` and other dependencies
 - run `poetry compose` to boot up the MongoDB. Erase any documents if present (as new attributes are required)
 - run `python data_transfer/main.py` and monitor the 'data > input and data > uploading' folder. _The code potentially fails if these folders are not present_

## To Do
I have not coded any tests yet, as I deemed it a priority to get this stage finished for now. Could possibly open a new Issue for this. But if required I am happy to spend a bit more time on writing tests. **Thoughts?**


data_transfer/lib/vttsma.py
```diff
@@ -0,0 +1,90 @@
+# NOTE: some methods exist for convience and will be advanced
+# for the CVS, e.g. serial_by_device_uuid might to hit a live 
+# endpoint rather than pull from a CSV file.
```
davidverweij, line 3: Just noticing some redundant comments. Will remove before merge.

data_transfer/lib/vttsma.py
```diff
@@ -0,0 +1,90 @@
...
+
+from data_transfer.config import config
+from data_transfer.services import dmpy
+from pathlib import Path
+from dataclasses import dataclass
+from typing import Optional
+from functools import lru_cache  # note: @cache exists in py >= 3.9
+import csv
```
davidverweij, line 11: I don't think we are using this anymore.. are we?

jawrainey: We do not -- it's been moved to `utils/__init__.py`

data_transfer/lib/vttsma.py
```diff
@@ -0,0 +1,90 @@
...
+    GET all records (metadata) from the AWS S3 bucket 
+
+    NOTE: S3 folder structure is symbolic. The 'key' (str) for each file object \
+        represents the path. See also `download_metadata()` in devices > vttsma.py 
+    """    
+    objects = bucket.objects.all()
+    object_paths = [obj.key for obj in objects]
+
+    # ignore users.txt files - data already present in object key
+    split_paths = [p.split('/') for p in object_paths if p.find('users.txt') == -1]
```
jawrainey, line 55: Since we know the format, I wonder if it would be more readable to create a dataclass such that we can do:

```python
split_paths = [DMPRecord(**p.split('/')) for p in object_paths if p.find('users.txt') == -1] 
```

Also, instead of using find, might we use `not in p`, e.g., 

```python
[p.split('/') for p in object_paths if 'users.txt' not in p]
```

data_transfer/lib/vttsma.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # follows [dump_date, raw/files, patienthash, patienthash.nfo/.zip/.audio?)]
+    # remove duplicates via a set()
+    patients = list(set([p[2] for p in split_paths]))
```
jawrainey, line 59: How come this is converted to a list?

data_transfer/lib/vttsma.py
```diff
@@ -0,0 +1,90 @@
...
+    
+    result = []
+    for patient in patients:
+        # TODO: only process dump dates of interest (i.e. since last run)
+        result.append(dict(id=patient, dumps=list(set([p[0] for p in split_paths if p[2] == patient]))))
+        
+    return result
```
jawrainey, line 66: Could be a list comprehension since you're building a new list and returning it?

```python
  return [
      dict(id=patient, dumps=list(set([p[0] for p in split_paths if p[2] == patient]))) 
      for patient in patients
  ]
```

data_transfer/schemas/record.py
```diff
@@ -10,11 +10,14 @@ class Record(BaseModel):
...
     id: Optional[Any] = Field(alias='_id')
     filename: str
+    device_type: str
     device_id: str
     patient_id: str
     start_wear: datetime
     end_wear: datetime
 
+    vttsma_dump_date: Optional[str]
+
```
jawrainey, line 20: Could this be `start_wear` and `end_wear` instead, since it stores similar same info? Should we be batching the SMA data per Dump Date rather than wear time (since dump date is within a wear time?)

data_transfer/services/dmpy.py
```diff
@@ -6,6 +6,13 @@ def zip_folder(path: Path) -> Path:
     return Path(shutil.make_archive(path, 'zip', path))
 
 
+def zip_folder_and_rm_local(path: Path) -> Path:
+    """Zips folder and removes the original immediately"""
+    zip_path = Path(shutil.make_archive(path, 'zip', path))
+    shutil.rmtree(path)
+    return zip_path
```
jawrainey, line 13: Maybe call the zip and rm methods above inside here instead?

data_transfer/services/ucam.py
```diff
@@ -65,6 +65,42 @@ def __device_from_record(device: Payload) -> Device:
...
+            vttsma_id=device.vttsma_id, 
+            devitations=device.devitations,
+            start_wear=device.start_wear,
+            end_wear=device.end_wear
+        )
+
+    devices = [__device_from_record(r) for r in patient_records]
+    return PatientRecord(patient, devices)
+
+
```
jawrainey, line 103: Maybe refactor this to a new method (__serialise_records) and call it here and in `get_record` since the code is 1-1?

jawrainey (commented):
Have tested locally and works correctly, including manually verifiying filenames against UCAM. I've added minor suggestions below on how the code might be improved, but otherwise it's perfect 👍🏼

data_transfer/devices/vttsma.py
```diff
@@ -0,0 +1,107 @@
...
+        # NOTE: currently downloads all dumps (inc. historical) TODO: only since last run
+        all_records = vttsma_api.get_list(self.bucket)
+
+        # Only add records that are not known in the DB based on stored filename (id = VTT hash id)
+        unknown_records = [r for r in all_records if r['id'] not in set(all_filenames())]
+
+        # Aim: construct valid record (metadata) and add to DB
+        for item in unknown_records:
+
+            if patient_record := ucam.record_by_vtt(item['id']):
```
jawrainey, line 67: Love it

data_transfer/devices/vttsma.py
```diff
@@ -0,0 +1,107 @@
...
+                device_used = [r for r in patient_record.devices if r.vttsma_id == item['id']]
+
+                # Assuming that only one device (phone) is used for the VTT SMA
+                device_used = device_used[0]
```
jawrainey, line 71: Ideally yes, only one patient will use one login credential (which we provide). However, some study sites have unfortunatly reused that once or twice between patients, so this 1-1 is not true for some historical data, and may not be true in the future. 

Maybe make a note here to edit this once we have real data on S3 to work against, e.g., from Newcastle?

data_transfer/devices/vttsma.py
```diff
@@ -0,0 +1,107 @@
...
+        Tracking: {db.record.is_downloaded} indicates success
+        
+        NOTE/TODO: is run as a task.
+        """
+        record = read_record(mongo_id)
+        is_downloaded_success = vttsma_api.download_files(self.bucket, record.filename, record.vttsma_dump_date)
+        if is_downloaded_success:
+            record.is_downloaded = is_downloaded_success
+            update_record(record)
+        # TODO: otherwise re-start task to try again
```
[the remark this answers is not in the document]
davidverweij: Yes definitely. I think it might be worth awaiting a few of the other devices, but can most likely be abstracted.

data_transfer/schemas/record.py
jawrainey: To add to this (having just thought about it), wouldn't the `Record` be the same if the patient has recorded data for two weeks in a row? So having the `end_wear` distinct would be a good way to determine between these Records

I suppose we could create multiple records with the same PatientID/DeviceID/WearPeriod, but do we want `Record` to be unique?

data_transfer/lib/vttsma.py
davidverweij: Good point. It's a remnant from when I used a list comprehension instead of a for loop (3 lines below).

data_transfer/devices/vttsma.py
davidverweij: Got inspiration from a certain person.....

data_transfer/schemas/record.py
davidverweij: Hmm good point, but currently the data within a dump date went beyond the dump dates itself. So I am not sure if that is a strong certainty we can rely on.

Also, as you noticed, VTT doesn't export date regularly, and perhaps we should not rely on that. We could, however, ignore all past 'dumps' if we store that somewhere. Then, we compute with all new dumps (most likely just one) and then this variable isn't needed anymore.

(I'm going to change the 'dump date' to 'export date' - keeps sounding silly as is)

davidverweij: I'm adding this to the Issue for separation and clarity. #23

data_transfer/devices/vttsma.py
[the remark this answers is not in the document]
davidverweij: Moved this to Issue #24

data_transfer/devices/vttsma.py
davidverweij: Added a TODO, as well as reference issue #25

data_transfer/services/ucam.py
davidverweij: Good point. Done and tested in this [commit](https://github.com/ideafast/middleware-services/pull/21/commits/cb462bc04765804a55acbfd9a5605f72315a2c96).

data_transfer/lib/vttsma.py
davidverweij: I've messed around with this for too long - and think it overcomplicates it, especially for now. I'll add my current thoughts below and push these into a separate issue for future work.

### schemas > vttsma_record.py
```python

from datetime import datetime
from dataclasses import dataclass


@dataclass
class VTT_patient:
    """
    A patient-based object constructed from VTT S3 bucket metadata
    """
    vttsma_hash: str
    export_dates: [str]

    def __init__(self, s3_bucket_entry: str)
        """
        extracting values from s3_bucket_entry.split following
        dump_date, raw/files, patienthash, patienthash.nfo/.zip/.audio? 
        """
        values = s3_bucket_entry.split('/')
        self.vttsma_hash = values[2]
        self.export_dates = [values[0]]

    def __eq__(self, other):
        return (
            isinstance(other, self.__class__) and
            getattr(other, 'vttsma_hash', None) == self.vttsma_hash
        )
```

### lib > vttsma.py > get_list()

```python
records = [SMA_Record(r.key) for r in objects if 'users.txt' not in r.key]
# TODO: create a set which removes duplicates, but merges the dump_dates
```

davidverweij: Tracking progress in Issue #26

davidverweij:
@jawrainey have implemented all comments as discussed - except for implementing the `SMA_Record` data class. I was running into challenges regarding converting a lists of paths (from the S3 bucket) into unique patient `SMA_Records` without too much logic. In the end my approach became too complex for what it is currently worth I think. I've set up issue #26 if we want to pursue this further.

Can you check my latest changes and approve/merge if appropriate?

jawrainey:
Great work @davidverweij 👍🏼 Tested locally and works for me as before

davidverweij:
@jawrainey , just putting it here before I forget. I see a lot of discussion online about AWS billing when interacting with AWS services and resources - especially by people new to AWS (like me!). Do we have an agreement with VTT on limits to their bucket? Just thinking ahead of preventing costs and potential headaches along the way.

Issue #36, opened 2021-02-05
Fix installing_with_constraints in .nox file


davidverweij:
Possibly a local issue, but @davidverweij cannot run `nox` commands fully as desired due to issues with `python`, `poetry` and `nox` combined. @jawrainey was unable to replicate these issues - so for now we circumvented the `install_with_constraints` in _noxfile.py_ by introducing:

```python
def tests(session: Session) -> None:
    """Setup for automated testing with pytest"""
    from cli import run_command

    run_command("pytest")
``` 
Ideally, this uses the `install_with_constraints()` wrapper to ensure replicate-ability across machines. I.e.:

```python
def tests(session: Session) -> None:
    """Setup for automated testing with pytest"""
    args = session.posargs
    session.run("poetry", "install", "--no-dev", external=True)
    install_with_constraints(session, "pytest")
    session.run("pytest", *args)
```

davidverweij:
Related to this is the use of `--without-hashes` which is needed for my machine, but again not being able to replicate elsewhere.

jawrainey:
@davidverweij -- wasn't this resolved in #37 and so can be closed?

davidverweij:
Unfortunately no, especially the `pytest` / `tests` setup. I'll try to investigate again soon.

Issue #35, opened 2021-02-05
Resolve typing complexity in tests and inventory


davidverweij:
Due to the unknown payloads from the inventory API, typing becomes more complex. As such, the response can return `Any` type:

```python
async def response(path: str, params: str = None) -> Any
```

This is further complicated in tests with the use of a monkeypatch, e.g. in _tests > consumer_tests > routers > conftest.py_:

```python
@pytest.fixture(scope="function")
    def response_row():
    ...
```
becomes something like
```python
@pytest.fixture(scope="function")
    def response_row() -> Generator[Dict[str, Any], None, None]:
    ...
```
This however fails typing as unkown methods are called on the Generator (e.g. `client.get()`) - which is used with the `monkeypatch`. Not to mention the 'private' methods in those tests and the required typing for their returned values.

At this moment, I've resorted to exclude these tests from typing checks by mypy, i.e. in `mypy.ini`:
```python
[mypy-consumer_tests.routers.conftest]
ignore_errors = True

[mypy-consumer_tests.routers.test_inventory]
ignore_errors = True
``` 

I've also excluded `flake8` linting for error code `AAA05` as it conflicts with `black` introducing a white line between private methods in an Act block.

### Next steps
Ideally this typing is corrected to prevent any coding errors, but priority is low.

PR #37, opened 2021-02-05
Implement test automation (nox) and linting and typing through pre-commit. Resolves #22

### LINKED ISSUE #22: Linting and Type Checking
To keep the codebase consistent and force standard practices, we should use the following:

Add the following as [dev dependencies](https://github.com/ideafast/dmpy/blob/master/pyproject.toml#L12-L20): 

- [x] [black](https://github.com/psf/black) (code formatter)
- [x] [Flake8](https://flake8.pycqa.org/en/latest/) (linting) with [relevant plugins](https://github.com/ideafast/dmpy/blob/master/noxfile.py#L44-L47)
- [x] use mypy to type check
- [x] pre-commit hook to automatically run linting and typing **(not testing as this will slow down commits)**
- [ ] noxfile to automate running the above via a CLI in a future PR 

See [dmpy](https://github.com/psf/black) for example of the above.

The above should then be applied to both the consumer and dtransfer packages.

davidverweij:
This PR implements linting, typing and tests such that upcoming contributions are sound - this should help with error prevention in the long run. 

More specifically, changes include (as suggested in #22):
- set up `nox` for test automation
- set up `nox` session to run:
   - `flake8` linting
   - `pytests` unit testing
   - `mypy` type testing
   - `isort` import statement ordering
   - `black` automatic code formatting
- set up `pre-commit` to run (and mandate passes) for these, except `pytests`. 

As expected, these prompted a breadth of required changes which are all resolved. A few simply cannot due to our setup, and have been added to config files for `mypy` and `flake8` to be ignored.

## Test
- pull down this branch
- run `poetry install` to include dev dependencies
- enter a virtual environment (`poetry shell`)
- set up pre-commit hooks: `pre-commit install`
- run individual check:
   - `poetry run nox -rs tests`. This runs all tests (and currently fails a few on the inventory part). Note that it currently circumvents `install_with_constraints_. See issue #36.
   - `poetry run nox -rs mypy`. This checks typing. Note that a few tests are currently excluded (see issue #35).
   - `poetry run nox -rs isort`. This sorts import statements alphabetically and by sections (i.e. global, 3rd party and local)
   - `poetry run nox -rs lint`. This runs various linting checks using flake8. Note that a few errors are ignored for our use case - see _mypy.ini_ in root.
   - `poetry run nox -rs black`. This checks **and changes** various formatting following _black_ standards.
- to test pre-commit functionality, make a simple change and commit. This should run (and pass) 
   - Check Yaml
   - Trim Trailing Whitespace
   - black
   - flake8
   - isort, and
   - mypy

## To Do
As referred to above, a few issues and todo's arose from the adoption of linting, tests and typing. These are documented in separate issues to work on.


davidverweij:
Not sure why, but `poetry run pre-commit run --all-files` reports a few more issues than any of the `nox` sessions. Possibly they are not picking up on config files..

.gitignore
```diff
@@ -9,3 +9,8 @@ __pycache__
 ideafast-*-full.csv
 *.json
 ucam_db.csv
+
+# mypy
+.mypy_cache/
+.dmypy.json
+dmypy.json
```
jawrainey, line 16: Can you ignore `.DS_Store` files and from the 3 from this PR (`**/.DS_Store`)? They're autogenerated by OSX

consumer/utils/errors.py
```diff
@@ -1,26 +1,29 @@
...
-from starlette.exceptions import HTTPException
 
 class CustomException(Exception):
     """
     Usage:
         raise CustomException(errors=["INVALID_TOKEN"], status_code=401)
     """
+
+    __slots__ = ["errors", "status_code"]
+
```
jawrainey, line 14: How come this magic attribute was added? [For reference](https://book.pythontips.com/en/latest/__slots__magic.html) re:what it does as I have not used it in a while.

data_transfer/lib/vttsma.py
```diff
@@ -1,67 +1,78 @@
...
+        aws_access_key_id=creds["aws_ak"],
+        aws_secret_access_key=creds["aws_ask"],
     )
-    
-    s3 = session.resource('s3')
-    bucket = s3.Bucket(creds['bucket_name'])    
-    
+
+    s3 = session.resource("s3")
+    bucket: Bucket = s3.Bucket(creds["bucket_name"])
```
jawrainey, line 21: Do you think we need to move `mypy_boto3_s3` as a dev dependency since it only exists for type checking?

data_transfer/devices/dreem.py
```diff
@@ -33,46 +33,51 @@ def download_metadata(self) -> None:
...
+                item["user"]
+            ) or inventory.patient_id_by_device_id(device_id)
+
             # NOTE: using inventory to determine intended wear time period.
-            # Useful for historical data, but (TODO) should be replaced with UCAM API. 
+            # Useful for historical data, but (TODO) should be replaced with UCAM API.
             history = inventory.device_history(device_id)[patient_id]
 
-            start_wear = utils.format_weartime(history['checkout'])
+            start_wear = utils.format_weartime(history["checkout"], "inventory")
```
davidverweij, line 65: @jawrainey , not that `mypy` indicated that a second argument was missing (hence I added "inventory" here and two lines further on). Do let me know if I misinterpreted this.

tests/consumer_tests/routers/conftest.py
```diff
@@ -19,13 +17,13 @@ def response_row():
...
     yield {
-        'asset_tag': 'SMP-TEST',
-        'model': {'name': 'TestName'},
-        'manufacturer': {'name': 'SmallCo'},
-        'status_label': {'status_meta': 'deployed'},
-        'location': {'name': 'Kiel'},
-        'serial': '119202',
-        'id': 1123,
+        "asset_tag": "SMP-TEST",
+        "model": {"name": "TestName"},
```
jawrainey, line 21: How come the folders within `/tests/` were renamed `/{package}_tests/`?

jawrainey (commented):
@davidverweij, Great job -- I have test the code and works as expected. I have also reviewed it and looks great to me besides the minor comment/questions below. 

Also, it seems the LGTM bot is not working as intended and getting confused on the merge, so no worries about that.

data_transfer/devices/dreem.py
jawrainey: It's correct and is a bug I introduced 👀  Thanks `mypy`!

.gitignore
davidverweij: Woops - an oversight on my end

noxfile.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+    session.run("flake8", *args)
+
+
+@nox.session(python=["3.8"])
+def tests(session: Session) -> None:
+    """Setup for automated testing with pytest"""
+    from cli import run_command
+
+    run_command("pytest")
```
jawrainey, line 74: Note: this is always a success regardless of return from pytest since it is not associated with the nox session (so cannot determine the value). For example, run:

```shell
poetry run nox -rs tests 
```

and then run `pytest` from the command line and note that 5 tests failed as `if config.access_authenticated_endpoints` is disabled.

Potential solution: 

change this code to `session.run("pytest", "-vs")` though not sure if that works for you @davidverweij

jawrainey: @davidverweij note: if `session.run` does not work for you locally, we could alternative use the result of `run_command` to notify if nox failed, e.g.,:

```python
@nox.session(python=["3.8"])
def tests(session: Session) -> None:
    """Setup for automated testing with pytest"""
    from cli import run_command
    result = run_command("pytest")
    if result.returncode:
        session.error("failed")
```

consumer/utils/errors.py
davidverweij: Flake8 / Mypy was alerting that we were using a Class which only sets its attributes in the ____init__.py__. It is suggested to, if it is immutable (and not used as a `dict` by adding attributes later) to either use a `namedTuple()` or `__slots__`. Adding this little line **fixes** the attributes, reducing memory and making it an immutable object.

data_transfer/lib/vttsma.py
davidverweij: My bad!

tests/consumer_tests/routers/conftest.py
davidverweij: `mypy` has issues when package resolve with similar names. There are complicated work-arounds to this, or a more simple one is to rename to avoid this issue. **thoughts** @jawrainey?

noxfile.py
davidverweij: Good catch!

It doesn't work, but `poetry run pytest -vs` works. Hence, temporary fix:
```python
@nox.session(python=["3.8"])
def tests(session: Session) -> None:
    """Setup for automated testing with pytest"""
    session.run("poetry", "run", "pytest", "-vs")
```

davidverweij: Can you test if that works for you @jawrainey?

jawrainey:
LGTM and `pytest` via nox now returns correct state

PR #41, opened 2021-02-12
separated validation method for tests, included tests

davidverweij:
Based on the ideafast/idgen, I've incorporated a IDEAFAST Device / Patient ID checker. This checks an ID (with study site prefix), and if valid - returns a formatted version (i.e. strips spaces and dashes). A few simple tests check the function, as well as the 'sub' method (which I had to pull out for testing) that replicates the `__get_remainder` copied from ideafast/idgen [here](https://github.com/ideafast/ideafast-idgen/blob/5200c6ca6fc9ac6c00f3f7d643ccf78293c03808/src/index.spec.ts#L18-L23).

### Test
In a `poetry shell`, run `poetry run nox -rs tests`


data_transfer/utils/__init__.py
```diff
@@ -47,3 +49,40 @@ def read_json(filepath: Path) -> Any:
...
+
+def validate_and_format_patient_id(ideafast_id: str) -> Union[bool, str]:
+    """
+    Validate a (messy) IDEAFAST id, based on ideafast/ideafast-idgen
+    Returns boolean if validate with corrected formatting
+    """
+
+    if type(ideafast_id) is str:
+        id_without_punc = re.sub(r"[^\w]", "", ideafast_id)
+        #  TODO: remove spaces if present
```
jawrainey, line 62: Potentially also remove tabs ("\t") that may be added historically to the inventory?

```python
id_without_punc = id_without_punc.replace(" ", "").replace("\t", "")
```

It would be worth adding another unit test to validate this too

jawrainey (commented):
Tests run and this will be very useful to have. I've made one minor comment re: removing spaces if present `TODO` that could be resolved in this PR?

data_transfer/utils/__init__.py
davidverweij: The `^\w` regex already identifies (and remove) spaces and tabs, but adding tests to prove that is a very good point.

davidverweij: Actually, it does include an underscore - I suppose we do want to remove those (however unlikely they are in there).  Then it should be:
```python
 id_without_punc = re.sub(r"[^\w]|_", "", ideafast_id)
```

jawrainey: Whoops -- My mistake! The `test_punctuation` test verifies no spaces, although potentially having an explicit test for tabs/spaces might be good anyway?

Issue #15, opened 2021-01-28
Add UCAM service


jawrainey:
We have been provided with a CSV export of the UCAM database. We should create a [service](https://github.com/ideafast/middleware-services/tree/master/data_transfer/services) for this that filters the data by device ID. I anticipate two potential endpoints that both return a modified `Record` (or similar):

```
GET /patients/  -> [Record]
GET /patients/<id>/ -> [Record]
```
As these endpoints don't exist in UCAM, we can instead load the CSV and filter it to mock the anticipated future implementation. We might have something like `history_by_device_id` and `history_by_patient_id` similar to the [inventory service](https://github.com/ideafast/middleware-services/blob/master/data_transfer/services/inventory.py).

PR #42, opened 2021-02-12
Use UCAM lookup in Dreem DAG

jawrainey:
> The current dreem data transfer service that uses the inventory should use UCAM instead once #15 implemented.

Previously, the code used the inventory API to determine which patients used which devices and when. This implementation now uses the UCAM API in addition to the inventory and simplifies the logic for how this is determined.

## Testing

1. Uncomment all but the metadata attribute in the [dreem DAG](https://github.com/ideafast/middleware-services/blob/master/data_transfer/main.py#L25-L35).
2. Configure `main.py` to auth and make queries to `newcastle`
3. Ensure mongoDB is running (`poetry run compose`) and drop the `records` collection.
4. Run the code and you'll see that a large portion of records is created for the data.

Note: I added print statements as logging as we can replace the `print` call with outputting to the log in the future and this is useful when running locally either way.


data_transfer/devices/dreem.py
```diff
@@ -46,41 +62,116 @@ def download_metadata(self) -> None:
...
+        return DreemRecording(*(id, device_id, user_id, start, end))
+
+    def __patient_id_from_ucam(
+        self, device_serial: str, dreem_start: datetime, dreem_end: datetime
+    ) -> Optional[str]:
+        """
+        Determine PatientID by wear period of device in UCAM.
+        """
+        # NOTE/TODO: given this is a 1-1 mapping, why not use a local CSV?
+        device_id = inventory.device_id_by_serial(device_serial)
```
jawrainey, line 158: Might be worthwhile to convert the `inventory.device_id_by_serial` API to read from a CSV since each device will be assigned one Serial ID and this is unlikely to change? This would save us making many external API calls to the inventory. Alternatively, we should look into caching the results from the inventory to prevent similar requests as this.

data_transfer/devices/dreem.py
```diff
@@ -46,41 +62,116 @@ def download_metadata(self) -> None:
...
+        record = ucam.record_by_wear_period(device_id, dreem_start, dreem_end)
+        # TODO: inventory has small rate limit.
+        time.sleep(2.5)
```
jawrainey, line 161: There is an API rate limit on the inventory of [150 requests by default](https://snipe-it.readme.io/reference#api-throttling). Critically, once the rate limit is met calls to the inventory's database appear to be disabled for a set period (about a minute) that would impact anyone using the inventory website.

We should increase this to a more reasonable limit, e.g., 3x current limit (450)? This requires modifying the source code within the docker container as there appears to yet be no way to do this via an `.env`.

data_transfer/devices/dreem.py
davidverweij: Agreed. With a number of `.csv` file we work from, might be good to add a folder in the repo.

data_transfer/devices/dreem.py
davidverweij: I got a connection error at metadata file nr 28 already - so quite a pressing issue. I am wondering if there are alternatives. Shot in the dark, but how about querying our past records (in the MongoDB) for some serials (if they are in there)? Should promptly return a result in recurring Dreem devices - which I expect is a common thing.

data_transfer/services/inventory.py
```diff
@@ -18,14 +20,27 @@ def device_history(device_id: str) -> Any:
...
-    if len(values) == 1:
-        return str(values[0]["patient_id"])
-    else:
-        # this means that there are multiple devices in the history, e.g.
-        # that multiple users may have used the same device. Therefore,
-        # we also need to pass in the wear time range and filter by it.
-        # TODO: check which device is within wear time range
-        return None
+    start_wear = utils.normalise_day(start_wear)
+    end_wear = utils.normalise_day(end_wear)
```
davidverweij, line 29: This is quite important, because are the days inclusive, or exclusive? `Normalise_day()` returns the day at 00:00, which means the end_wear is exclusive. 

I need to watch out for this too, as I was using `datetime.replace(hour=23, minute=59)` for the byteflies implementation. Could use a utils method for this!

For readability, I'd suggest to have two methods, `normalise_day_start` and `normalise_day_end`.

data_transfer/services/inventory.py
```diff
@@ -18,14 +20,27 @@ def device_history(device_id: str) -> Any:
...
+        checkin = record["checkin"] or datetime.now().strftime(
+            utils.FORMATS["inventory"]
+        )
+        inventory_end_wear = utils.format_weartime(checkin, "inventory")
+
+        inventory_start_wear = utils.normalise_day(inventory_start_wear)
+        inventory_end_wear = utils.normalise_day(inventory_end_wear)
+
+        within_start_period = inventory_start_wear <= start_wear <= inventory_end_wear
+        within_end_period = inventory_start_wear <= end_wear <= inventory_end_wear
```
davidverweij, line 42: I now see you use normalised days for comparison. I agree with your current implementation but must admit this doesn't feel fully accurate. Than again - our use case will never have hit or misses in these type of comparisons based on hours - and even not based on days! So all good here.

data_transfer/services/ucam.py
```diff
@@ -98,12 +98,38 @@ def record_by_wear_period(
...
+    If data was created on a certain period then it belongs to an individual patient.
+    """
+    return __record_in_wear_period(devices, start_wear, end_wear)
+
+
+def __record_in_wear_period(
+    records: List, start_wear: datetime, end_wear: datetime
+) -> Optional[Payload]:
+    """
+    Shared method given the logic is the same above
```
davidverweij, line 119: Nit-picky, but would change the description to be standalone.

data_transfer/services/ucam.py
```diff
@@ -98,12 +98,38 @@ def record_by_wear_period(
...
+    start_wear = normalise_day(start_wear)
+    end_wear = normalise_day(end_wear)
+
+    for record in records:
+        # While we could compare records to the second, this caused some issues, e.g.,
+        # when a record was created on the day for testing but not checked out until afterwards.
+        drm_start_wear = normalise_day(record.start_wear)
+        drm_end_wear = normalise_day(record.end_wear)
+
+        within_start_period = drm_start_wear <= start_wear <= drm_end_wear
```
davidverweij, line 130: Although this method is currently solely used for Dreem, I would argue to keep the naming device agnostic. I.e. `drm_start_wear` could be `record_start_wear`

data_transfer/services/inventory.py
davidverweij: I now see your use is different from mine for the ByteFlies implementation. So you can ignore this suggestion.

davidverweij (commented):
Nice one @jawrainey. I have a few comments regarding naming, though predominantly thoughts I want to share with you for future improvements.

If you can respond to the comments (and just resolve them as you have read them) - I'd be happy if you could merge the branch to master.

data_transfer/devices/dreem.py
davidverweij: Actually, how many Dreem devices do you foresee we having in our inventory? Otherwise, we could alternatively add another query in the  _consumer > routers > inventory_, such as `/device/bytype/{type}` with type being the Dreem device. 

At the start of the Dreem DAG, this would the query `snipeit` once, e.g.
```shell 
GET snipeit_url/hardware?limit=500&model_id=8'
```
where model = 8 corresponds to the Dreem Device I believe. If you think more than 500 Dreem devices might be present, we could increase the number of items by increasing [MAX_RESULTS in our .env file](150 requests by default). We could even further filter the results by adding a `location_id` parameter Thoughts?

data_transfer/devices/dreem.py
```diff
@@ -46,41 +62,116 @@ def download_metadata(self) -> None:
...
+            # Serial may not exist in lookup, e.g., if Dreem send a device replacement.
+            if not device_serial:
+                print(f"Unknown Device:\n   {recording}")
+                # Move onto next record: skips logic below to simplify error handling
+                continue
+
+            # Record before use in next loop iteration
+            record = None
+            # Used to filter UCAM devices and assign to type to record
+            dtype = utils.DeviceType.DRM.name
```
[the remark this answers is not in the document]
jawrainey: Great suggestion -- have updated it and renamed it to `device_type` to be explicit

data_transfer/devices/dreem.py
```diff
@@ -8,7 +11,20 @@
...
+class DreemRecording:
+    """
+    Stores most relevant metadata for readable lookup.
+    """
+
+    id: str
+    device_id: str
+    user_id: str
+    start: datetime
+    end: datetime
```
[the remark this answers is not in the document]
jawrainey: In short, I have used the dataclass primarily for creating nice output to the logs otherwise the code becomes a bit of a mess. For reference, this is because dataclasses implement `[__repr__](https://docs.python.org/3/reference/datamodel.html#object.__repr__)` quite well. This outputs:

```json
Metadata cannot be determined for:
    DreemRecording(id='HASH_ID', device_id='HASH_ID', user_id='HASH_ID', start=datetime.datetime(2020, 12, 9, 21, 35), end=datetime.datetime(2020, 12, 10, 5, 38, 23)
```

> I understand the use of a data class here, and it helps in making its usages more strict whilst coding. Personally, I am not sure if I would use it in this case - as it is only used in the for item in unknown_records loop. As the dict is constructed via a 'private' method - it might actually make the translation between the Dreem API payload and our use a bit less transparent.

Note that `__recording_metadata` is essentially unpacking specific dictionary values from the complete metadata file, i.e., only the ones of interest. For example, previously the following was in the code:

```python
# Instead of using a dataclass:
recording = self.__recording_metadata(item)

# I had written:
record_id = recording["id"]
record_device_id = recording["device"]
record_user_id = recording["user"]
record_start = datetime.fromtimestamp(recording["report"]["start_time"])
record_end = datetime.fromtimestamp(recording["report"]["stop_time"])
```

Note that _logging_ takes place at multiple locations throughout this class, and critical to this is outputting all the above information. As sucess, previously we may have written:

```python
if not device_serial:
    # this would be duplicated throughout the code
    print(f"Unknown Device:\n   RecordingID: {record_id}, DeviceID: {record_device_id}")
    # Move onto next record: skips logic below to simplify error handling
    continue
```

With dataclasses, this becomes:

```python
if not device_serial:
    print(f"Unknown Device:\n   {recording}")
    # Move onto next record: skips logic below to simplify error handling
    continue
```

I agree that the intended use of this dataclass could be clearer. Shall I update a comment in the docstring and use of this? Also, do you agree/like the idea of using dataclasses in this way? This could be something that we do across all devices if desired.

data_transfer/devices/dreem.py
```diff
@@ -46,41 +62,116 @@ def download_metadata(self) -> None:
...
         ]
+        known, unknown = 0, 0
 
-        # Aim: construct valid record (metadata) and add to DB
         for item in unknown_records:
+            # Pulls out the most relevant metadata for this recording
+            recording = self.__recording_metadata(item)
+
+            # There is be a 1-2-1 mapping between IDs and serials via CSV lookup.
+            device_serial = dreem_api.serial_by_device(recording.device_id)
```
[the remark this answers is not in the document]
jawrainey: I agree that we should pull out the use of CSV methods like `__key_by_value()`. The intention of `serial_by_device` is to "mock" a potential future API by dreem, e.g., we could replace this method with an API call in the future if it exists. So in this regard I would argue for not moving the `dreem_serial_by_device` to `csv_lookups`. Nevertheless, I do agree that we should a `csv_lookups` module/file such as for UCAM as you've suggested 👍🏼

data_transfer/devices/dreem.py
```diff
@@ -46,41 +62,116 @@ def download_metadata(self) -> None:
         unknown_records = [
             r for r in all_records if r["id"] not in set(all_filenames())
         ]
+        known, unknown = 0, 0
```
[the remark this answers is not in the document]
jawrainey: I agree -- this was mostly for a local sanity check as it verified that the logic was correct, but at the same time is useful logging information. The question is: *do we want to store this information inside our classes, or simply generate it via a summarisation of logs (of which external software exists to do that)*? 

I agree that we want consistency between logging of devices, and something we should discuss in-depth 👍🏼

data_transfer/devices/dreem.py
jawrainey: > Actually, how many Dreem devices do you foresee we having in our inventory? Otherwise, we could alternatively add another query in the consumer > routers > inventory, such as /device/bytype/{type} with type being the Dreem device.

Currently **46** and potentially at most **200** in the future.  Great find @davidverweij -- not sure how I missed it! We might be able to filter by `serial` as a GET parameter too -- I will have a look. 

> If you think more than 500 Dreem devices might be present, we could increase the number of items by increasing [MAX_RESULTS in our .env file](150 requests by default). We could even further filter the results by adding a location_id parameter Thoughts?

The bigger issue is that even though we have 40 devices, we could make potentially 3 API calls to inventory per data record, so we would hit the threshold of 150 requests quite easily. A caching solution for the same queries would be another potential solution since the queries are often the same. For example, for dreem if there are 10 data recordings created on the same device, we might have to lookup the DeviceID by serial 10 times.

data_transfer/devices/dreem.py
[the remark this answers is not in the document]
davidverweij: Thanks for the elaborate response Jay. Yes, it does make (more) sense now, especially with the simplicity for logging. 

How would you then (for example) approach the VTT dataclass? The API query only returns a limited set of attributes for the data class (compared to the Dreem, above). Basically only the `record_id` (which is the `vtt_hash`). Would you then update dataclass attributes as we uncover them, or only use the dataclass object for information directly from the manufacturer's API?

data_transfer/devices/dreem.py
jawrainey: Interesting ... not sure if the connection error is related (might be MongoDB?), but either way is a pressing issue.

> Shot in the dark, but how about querying our past records (in the MongoDB) for some serials (if they are in there)? Should promptly return a result in recurring Dreem devices - which I expect is a common thing.

We do not store serials in mongoDB although we might for convience to resolve this issue? Alternatives could be: (1) store serial->ID mapping in a CSV; (2) cache the API request; (3) store serials in MongoDB and do a lookup there.

Perhaps (3) is reasonable to store under the "meta" field?

data_transfer/devices/dreem.py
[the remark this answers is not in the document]
jawrainey: > How would you then (for example) approach the VTT dataclass? The API query only returns a limited set of attributes for the data class (compared to the Dreem, above). 

In VTT's case we could have `id` and `date` (if we plan to use it)? In this case, a dataclass may feel inappropriate since it only has 1 or 2 values, but for consistency with other devices and logging it may be useful. 

> Basically only the record_id (which is the vtt_hash). Would you then update dataclass attributes as we uncover them, or only use the dataclass object for information directly from the manufacturer's API?

*I use the dataclass as a way to serialise the manufacturer's API to have more meaningful naming to make the code cleaner to read.*

Take Byteflies for example, you would have multiple values: `startDate`, `duration`, `id`, `groupId`, etc. The purpose of the data class was to create more meaningful names for these attributes. In the dreem implementation, the manufactuer's response has _a lot_ of metadata. We want to store it all, so although I construct the `DreemRecording` the complete object is not stored elsewhere and instead `item` is [saved](https://github.com/ideafast/middleware-services/blob/ft_ucam_dreem/data_transfer/devices/dreem.py#L132).

> If we do update the dataclass attributes as we uncover them, it is basically replacing the Record object. So if that is o.k. - we could also have the __recording_metadata() translate the Dreem payload straight into a Record object. thoughts?

This is a good point, however, the use of the dataclass is not _replacing_ the `Record` object but instead serialises the response from the manufacturer with only the relevant data needed to construct a Record. This makes (imho) reading the code easier, which is the key use of the dataclass here. Also note that the `item` metadata (manufactuer's response) is stored in `-meta.json` and not the dataclass.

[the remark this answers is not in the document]
davidverweij: Got it, that makes senses. I'd say when we implement these across devices they'll become more obvious. Perhaps a few naming improvements would distinguish/clarify them a bit more - but overall I agree with your reasoning and the use of the different classes. Thanks for the elaborate explanation!

data_transfer/devices/dreem.py
davidverweij: I've referenced this discussion on the [issue on the stack](https://github.com/ideafast/stack/issues/9#issuecomment-781258338)

data_transfer/devices/dreem.py
davidverweij: I've referenced this discussion on the issue one [the stack](https://github.com/ideafast/stack/issues/9#issuecomment-781258338)

Issue #34, opened 2021-02-04
Update MongoDB record structure and pipeline protocol


davidverweij:
From @jawrainey: given we're using mongodb, we could have a "meta" attribute on the Record that stores relevant metadata in json format for specific devices that we can access as needed? 
 
That way, we don't have to continue adding device specific attributes

davidverweij:
In light of this - I am still a bit worried about the use of the `uid`s (provided by the manufacturer) as a unique identifier (not `_id` but `filename`) to check for past downloads in the MongoDB. If we need this logic, I'd suggest to add a prefix to ensure separation of manufacturer uids. In addition these filenames link to specific files/records to download, and not necessarily for patient identifiers. I am unfamiliar how this is with Dreem - but it already poses an issues with VTT SMA if a patient occurs across data exports (and we keep mongoDB records live). 
### Suggestion
My suggestion then would be to add the `DeviceType` enum name value as prefix, such that we get things like:
- AX6-...
- DRM-...
- SMA-...
...

For BTF, each record has a uid. Each signal data has their own unique ID as well, thus we can use that for uids (given the prefix) in the mongoDB. 

For VTTSMA I would tightly couple the data export name to the mongoDB entry, as a patient hash might occur across data exports, but not if we include the wear period - as VTT already does in their exports. Thus the monogDB filename would be something akin to `SMA-qwertyuiop-01012020-02022020`. If we fear that we might get data exports with identical wear times and patient hashes across data exports - we might as well include the full pathname, i.e. `SMA-data_export_01012020-qwertyuiop-01012020-02022020`.

jawrainey:
@davidverweij -- for dreem the `filename` is distinct as data generated on a device is unique. The mapping between patients who used the device's based on the `device_id` was a bit tricky for historical reasons. I can see that for VTT-SMA the filename hash may be duplicated given the application may have been logged in by multiple patients.

What's the benefits of storing the filename in the format above above over using the existing properties of `Record`, which  store this metadata, i.e., [`device_type`](https://github.com/ideafast/middleware-services/blob/master/data_transfer/schemas/record.py#L15) and the wear period ([`start_wear`](https://github.com/ideafast/middleware-services/blob/master/data_transfer/schemas/record.py#L18) and [`end_wear`](https://github.com/ideafast/middleware-services/blob/master/data_transfer/schemas/record.py#L19)?

[...]

PR #43, opened 2021-02-23
ByteFlies pipeline

davidverweij:
This is a draft pull request revolving the ByteFlies pipeline. Due to the ByteFlies having multiple files to download, this PR includes multiple changes that break the other pipelines (for as far VTT was already broken). This is due to a number of required changes (as discussed in #34) in order to get this to work. 

## A rough list of most important changes
See a more detailed discussion about some of these in Issue #34 
- A `Record` now has a hash instead of filename 
- A `Record` stores a `manufacturer_ref` for referencing all records (and their files) to a single recording - per manufacturer. I.e., ByteFlies has ~10 files per recording, where `manufacturer_ref` links to the `id` of that recording for all.
- the `-meta.json` file relates to a recording (depending on manufacturer). A `-preprocessed.json` file can be generated **per file / Record**. This is a clear distinction.
- files (and `-meta.json` files) are now downloaded in _input/PATIENT_ID/DEVICE_ID_ , instead of flattened in _input_. Then, in `task_prepare_data()` these are translated into _PATIENTID-DEVICEID-START-END_ for DMP. If a folder is not processed (i.e. unfinished), new files can easily be added to existing folders 
- `shared. task_prepare_data()` is abstracted to function across device pipeline (`shared_jobs.prepare_data_folders()`). It checks for (sub)folders in the `input` folder, and proceeds to package them for DMP upload if they are ready to be uploaded (i.e. all Records related to the folder have been successfully downloaded and pre-processed).
   -  note the additional methods in `data_transfer/db/main/py` that allow retrieving a list of unfinished 'folders', a list of `Records` coupled to an _input_ folder, and the max and min dates of data across those records.
- A `patient_id` and `device_id` validator is available in the `utils`. Note that this removes any punctuation (e.g. `- dashes`, and cannot be used in - for example - the inventory call (which stores `device_id`s with dashes). 
- For ByteFlies, their API returns a list of recordings, which need to be individually called to get a full list of files to download. This API call is re-run in the download stage to get the latest download URL.
- The API payload of BTF is deep copied (using `__metadata_copy`) in order to create a Record in the next stage. However, only one `-meta.json` file remains for each recording - even though this metadata is deep copied as many times as there are files to be downloaded for each recording.
- `__get_response()` in _lib/byteflies.py_ has a `timer.sleep(1)` to reduce call to their API. 
- A hash is created from information that makes the recording/file unique. For BTF this is the `recording_id`/`signal_id` and if needed /`algorithm_id`. This hash is temporarily stored in the individual result from `lib.get_list()` and removed before writing the `-meta.json`.
 

### To test
- place the `ideafast-byteflies-devices-full.csv` from Teams in the root folder
- update the `.dtransfer.env` with the new/added credentials for Byteflies (see KeePass)
- comment [the inventory endpoints if statement](https://github.com/ideafast/middleware-services/blob/895e57cc650c85fc5f3eee9374abe40fba51119f/consumer/main.py#L14) (and undo indents)
- run `poetry install`
- run `poetry shell`
- run `poetry run consumer` to get the inventory API up an running
- in another terminal window:
- run `poetry shell`
- run `poetry run compose` to get the MongoDB up and running
- run `python data_transfer/main.py` and see metadata being downloaded, files being downloaded, and being moved into the upload folder.

The pipeline only gets [the latest 2 days](https://github.com/ideafast/middleware-services/blob/895e57cc650c85fc5f3eee9374abe40fba51119f/data_transfer/main.py#L79) - which you can adjust. It does however look across study sites at the moment.

You can test if the `prepare_data_folder()` stage ignores 'unfinished' folders by commenting the [last two tasks](https://github.com/ideafast/middleware-services/blob/895e57cc650c85fc5f3eee9374abe40fba51119f/data_transfer/main.py#L87-L90) - and then edit one of the files in the MongoDB and set `is_processed = False`. Then comment the `download` and `preprocess` tasks and uncomment the `prepare_data_folders`.

### Questions / considerations 
The main consideration I am hesitant about it the current approach in `prepare_data_folder()`. Even though it checks that all `Records` associated with a folder are completed, it assumes all files (linked to individual `Records`) are in that folder. 

The alternative would be to get each file associated to a `Record` (i.e. the `manufacturer_ref-meta.json`, `signal or algorithm_id.csv` and possibly a `signal or algorithm_id-preprocessed.json` files) and check if they are present. However, then we would need to track if a `preprocessed.json` is created, and what the files names were once created - which different per device implementation.

**thoughts?**

### Todo (before merging or at some point)
- [x] transform ByteFlies pipeline to run study sites separately
- [ ] introduce a [Inventory endpoint that returns a list of devices per deviceType](https://github.com/ideafast/stack/issues/9#issuecomment-781258338) - this should/could replace the serial .csv lookup (for both Dreem and ByteFlies).
- [ ] clean ByteFlies pipeline, i.e
    - [ ] remove empty download folder in _input_ 
    - [ ] cache ByteFlies API calls (when downloading files from the same recording - should only be available very briefly!)
- [ ] update Dreem pipeline to adapt to these changes
- [ ] update VTT pipeline to adapt these changes
- [x] potentially split validation and formatting functionality of `format_patient_id` for more flexible use.


data_transfer/jobs/shared.py
```diff
@@ -26,3 +45,68 @@ def batch_upload_data() -> None:
...
+    to_process = [
+        taskObject(
+            folder[0],
+            folder[1],
+            (config.storage_vol / folder[0] / folder[1]),
+            get_records_in_folder(folder[0], folder[1]),
+        )
+        for folder in ready_folders
+    ]
+
```
jawrainey, line 84: As discussed elsewhere, this could be simplified by working with the database directly and removes a need to read folder structures that may change. 

Instead, I would work with the DeviceID and PatientID since they create a composite key that represents the files to be uploaded. In this way, we don't need to read any files or look in folders. To do that, I would query the DB such that it returns a list of all records where `is_downloaded, is_processed` are True, i.e, they need to be uploaded but have not been downloaded?

I would then transform that output to the following as it represents each Record in each folder

```python
{
  PatientID1-DeviceID1: [{Record1}, ... {Record2}],
  PatientID1-DeviceID2: [{Record1}, ... {Record2}]
}
```

From this, you can then filter this dict of lists as follows: if `is_downloaded=True` is in any `Record`. That way, now your final dict contains all PatientID/DeviceIDs that have a list of fully downloaded files and are thus ready to upload.

data_transfer/lib/byteflies.py
```diff
@@ -0,0 +1,227 @@
...
+@dataclass
+class BytefliesFileDownload:
+    """Use as CLI arguments for ByteFlies' library."""
+
+    directory: Path = config.storage_vol
+    ftype: str = "csv"
+
+
+# Define location and filetype to download
+args = BytefliesFileDownload()
```
jawrainey, line 22: This can be removed since they're always CSV files?

data_transfer/lib/byteflies.py
```diff
@@ -0,0 +1,227 @@
...
+
+def __download_file(download_folder: str, url: str, filename: str) -> None:
+    """
+    Builds the target filename and starts downloading the file to disk
+    NOTE: Download folder is always /PATIENT_ID/DEVICE_ID
+    """
+    folder = Path(config.storage_vol) / download_folder
+    if not folder.exists():
+        folder.mkdir(parents=True, exist_ok=True)
+    file_path = folder / f"{filename}.{args.ftype}"
```
jawrainey, line 221: change `args.ftype` to `.csv` as above?

data_transfer/main.py
```diff
@@ -50,13 +53,44 @@ def vttsma_dag() -> None:
...
+
+    for record in records_not_downloaded(DeviceType.BTF):
+        # Each task should be idempotent. Returned values feeds subsequent task
+        mongoid = byteflies_tasks.task_download_data(record.id)
+        mongoid = byteflies_tasks.task_preprocess_data(mongoid)
+
+    # Data is finalised and moved to a folder in /uploading/
+    shared_jobs.prepare_data_folders()
     # All said folders FOR ALL DEVICES are uploaded once per day
     shared_jobs.batch_upload_data()
```
jawrainey, line 90: As noted above, we likely want this on a per-device rather than `shared_jobs`

data_transfer/lib/byteflies.py
```diff
@@ -0,0 +1,227 @@
...
+            copy["IDEAFAST"] = {
+                "hash": uid_to_hash(
+                    f"{copy['id']}/{signal_id}/{algorithm_id}", DeviceType.BTF
+                ),
+                "signal_id": signal_id,
+                "algorithm_id": algorithm_id,
+            }
+            return copy
+
+    groups = __get_groups(session)
```
jawrainey, line 84: might be worth renaming `groups` here and elsewhere to `study_site`. Although they're groups on byteflie's platform, it's easier to think of them as study sites?

jawrainey (commented):
> The main consideration I am hesitant about it the current approach in prepare_data_folder(). Even though it checks that all Records associated with a folder are completed, it assumes all files (linked to individual Records) are in that folder.

As mentioned above, it would be ideal instead to work with the database directly and do this per-device rather than for all devices. Likewise, perhaps `batch_upload_data` should be modified to also work per-device?

data_transfer/jobs/shared.py
```diff
@@ -26,3 +45,68 @@ def batch_upload_data() -> None:
...
+            config.upload_folder.mkdir()
+
+        if not destination.exists():
+            destination.mkdir()
+
+        # assuming all downloads went well: move everything and update records
+        for fname in task.folder.glob("*.*"):  # grabs all files
+
+            new_path = destination / fname.name
+            fname.rename(new_path)
```
jawrainey, line 107: This also means you can remove this code, since we should move the complete folder (and all its contents) from `/input/PatientID/DeviceID` to `/upload/{destination}`

data_transfer/jobs/shared.py
```diff
@@ -26,3 +45,68 @@ def batch_upload_data() -> None:
...
+    ]
+
+    for task in to_process:
+        max_data, min_data = min_max_data_wear_times(task.records)
+
+        start_data = max_data.strftime("%Y%m%d")
+        end_data = min_data.strftime("%Y%m%d")
+
+        data_folder_name = (
+            f"{task.patient_id}-{task.device_id.replace('-', '')}"
```
jawrainey, line 92: We should instead save the folders in `input/` with valid DeviceIDs for the DMP, i.e., without "-". This can then be removed

data_transfer/devices/byteflies.py
```diff
@@ -0,0 +1,222 @@
...
+            )
+
+            create_record(record)
+            print(f"Record Created:\n   {record}")
+
+            del item["IDEAFAST"]  # remove generated temporary metadata
+            # NOTE: this gets overwritten for each to-download file = most recent version
+            # Could argue to pass if already exists (= oldest version)
+            path = Path(config.storage_vol / f"{record.manufacturer_ref}-meta.json")
+            utils.write_json(path, item)
```
jawrainey, line 141: Are the contents of the file the same across devices? If yes, then your argument for passing it if it already exists  is great and we should do that 👍🏼

data_transfer/db/main.py
```diff
@@ -32,10 +32,42 @@ def update_record(record: Record) -> None:
...
+def get_records_in_folder(patient_id: str, device_id: str) -> List[Record]:
+    docs = _db.records.find(
+        {
+            "device_id": device_id,
+            "patient_id": patient_id,
+            "is_processed": True,
+            "is_prepared": False,
+        }
+    )
+    return [Record(**doc) for doc in docs]
```
jawrainey, line 63: Given proposed update, this likely needs removed?

data_transfer/jobs/shared.py
```diff
@@ -26,3 +45,68 @@ def batch_upload_data() -> None:
...
+
+
+def prepare_data_folders() -> None:
+    """
+    Checks folders present in 'data/upload/ are finished
+    and moves them into the upload folder in the format:
+
+        DEVICEID-PATIENTID-STARTWEAR-ENDWEAR
+    """
+
```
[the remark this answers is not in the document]
davidverweij: I've tried to come up with a device-specific `batch_upload_data`, which is either:
- add a `regex` to determine which folders include the device acronym
- move folders to upload into device specific folders, e.g. `uploading/BTF/`

I find the first dodgy, the second acceptable. Do we want to do this in order to stage the upload better? Or leave as is (upload for all devices) and just make sure pipelines are executed sequentially?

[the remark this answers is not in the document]
jawrainey: > I find the first dodgy, the second acceptable. Do we want to do this in order to stage the upload better? Or leave as is (upload for all devices) and just make sure pipelines are executed sequentially?

Yeah, might be better to have them separate as pipelines will likely run independently. Another option could be to filter folders before zipping based on device type since it's in the folder name and would be passed as parameter (in the none-sequential, device specific `batch_upload_data` method)?

```python
folders_to_upload = [p for p in config.upload_folder.iterdir() if p.is_dir() and NAME in p]
```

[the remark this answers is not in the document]
davidverweij: Yes, sorry - that's what I meant with the `regex`. I just felt that it was slightly dodgy - as the acronym might just possibly be in the device_id or patient_id (although very very unlikely). But happy to include if you feel that is just never going to happen!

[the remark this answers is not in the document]
jawrainey: Great point 👍🏼🚀  You're right, option two (`uploading/device_id`) would be ideal and likely simplifies the code in prepare too

davidverweij:
I've refactored the code to prepare folders for upload based on a DB query - which works! I've also amended any suggestions you had, but still some polishing to do. Will run another test next week (I was trying to set some tests up... - haven't gotten far) and then remove the 'Draft' from this PR

davidverweij:
I've rebased this branch with master, included some tests (Work in Progress). The data download and prepare not appropriately structure folders with multiple files per downloaded recording.

## Testing
- place the ideafast-byteflies-devices-full.csv from Teams in the root folder
- update the .dtransfer.env with the new/added credentials for Byteflies (see KeePass)
- comment the [inventory endpoints if statement](https://github.com/ideafast/middleware-services/blob/895e57cc650c85fc5f3eee9374abe40fba51119f/consumer/main.py#L14) (and undo indents)
- run poetry install
- run poetry shell
- run poetry run consumer to get the inventory API up an running

in another terminal window:
- run poetry shell
- run poetry run compose to get the MongoDB up and running
- run python data_transfer/main.py and see metadata being downloaded, files being downloaded, and being moved into the - upload folder.

data_transfer/devices/byteflies.py
```diff
@@ -0,0 +1,227 @@
...
+        ...
+
+        NOTE/TODO: will run as BATCH job.
+        """
+        # Note: includes metadata for ALL data records, therefore we must filter them
+        all_records = byteflies_api.get_list(
+            self.session, config.byteflies_group_ids[study_site], from_date, to_date
+        )
+        # utils.write_json("./byteflies.json", all_records)
+        # all_records = utils.read_json(Path("./byteflies.json"))
```
jawrainey, line 84: Remove?

data_transfer/devices/byteflies.py
```diff
@@ -0,0 +1,227 @@
...
+                    signal_id=recording.signal_id,
+                    algorithm_id=recording.algorithm_id,
+                ),
+            )
+
+            create_record(record)
+            print(f"Record Created:\n   {record}")
+
+            del item["IDEAFAST"]  # remove record-specific temporary metadata
+            path = Path(config.storage_vol / f"{record.manufacturer_ref}-meta.json")
```
jawrainey, line 142: Note: this stores the "meta.json" inside `/input/` rather than `/input/{resolved_patient_id}/`. At this point then, it would be worth making the folder `/{resolved_patient_id}/{resolved_device_id}/`

Perhaps a helper method (in `utils.py`) as this will be used in `download_file` and the same for all pipelines?

data_transfer/lib/byteflies.py
```diff
@@ -0,0 +1,225 @@
...
+) -> None:
+    """
+    Builds the target filename and starts downloading the file to disk
+    NOTE: Download folder is always /PATIENT_ID/DEVICE_ID
+    """
+    folder = Path(config.storage_vol) / download_folder
+    if not folder.exists():
+        folder.mkdir(parents=True, exist_ok=True)
+    file_path = folder / f"{filename}.csv"
+    response = session.get(url, stream=True)
```
jawrainey, line 220: Note: when using `Stream=True` we should call this assignment within a [with statement](https://requests.readthedocs.io/en/master/user/advanced/#body-content-workflow) to ensure the contet is closed. That would mean the reading file bit below would be intended within this with similar to how I did it in the latest [dreem branch](https://github.com/ideafast/middleware-services/compare/docker_pipeline#diff-666c9d8f23463c4da86839196dc51f9d3b26e0e36d7a3e05bdf0cce458f13814R139-R140).

data_transfer/utils/__init__.py
```diff
@@ -23,13 +24,40 @@ class DeviceType(Enum):
     YSM = 11  # ZKOne YOLI
 
 
+class StudySite(Enum):
+    Newcastle = 1
+    Kiel = 2
+    Muenster = 3  # i.e. Münster
+    Rotterdam = 4  # i.e. Erasmus
```
jawrainey, line 31: Great idea! 🥳

data_transfer/jobs/shared.py
```diff
@@ -26,3 +40,57 @@ def batch_upload_data() -> None:
...
+        if not destination.exists():
+            destination.mkdir(parents=True)
+
+        source.rename(destination)
+
+        # don't forget the -meta.json file! Each record in the list has a reference to it
+        meta_file_name = f"{to_upload[patient_device][0].manufacturer_ref}-meta.json"
+        meta_file_path = config.storage_vol / meta_file_name
+        meta_file_path.rename(destination / meta_file_name)
```
jawrainey, line 86: This could be removed if `-meta.json` files are in the `source` folder. Same is true for all pre-processed files.

data_transfer/jobs/shared.py
```diff
@@ -26,3 +40,57 @@ def batch_upload_data() -> None:
...
+        max_data, min_data = min_max_data_wear_times(to_upload[patient_device])
+
+        start_data = max_data.strftime("%Y%m%d")
+        end_data = min_data.strftime("%Y%m%d")
+
+        source = config.storage_vol / patient_device
+        destination = (
+            config.upload_folder
+            / device_type.name
+            / f"{patient_device.replace('-','').replace('/','-')}-{start_data}-{end_data}"
```
jawrainey, line 75: For consistency, might we want to have the upload folder mirror the input folder, e.g., to be `/{patient_id}/` rather than `/{deviceType}/{patient_id}/`? If yes, then device_type.name can be changed to `patient_device.split("/")[0]`

This related specifcally to line 74: -> `/ device_type.name`

data_transfer/jobs/shared.py
```diff
@@ -26,3 +40,57 @@ def batch_upload_data() -> None:
...
+
+        for record in to_upload[patient_device]:
+            record.is_prepared = True
+            update_record(record)
+
+        # check if patient folder is empty, then remove it
+        patient_path = config.storage_vol / to_upload[patient_device][0].patient_id
+        if not any(patient_path.iterdir()):
+            # throws if not empty
+            patient_path.rmdir()
```
jawrainey, line 96: Not sure why, but this doesn't seem to get called when I tested it just now.

jawrainey (commented):
LGTM besides minor points below 🚀

data_transfer/devices/byteflies.py
davidverweij: Good suggestion. I've added this as a helper method for the `Record` class:

```python
    def download_folder() -> str:
        """consistenly structure target folder for download"""
        return f"{device_type}/{self.patient_id}/{self.device_id}"
```

data_transfer/jobs/shared.py
davidverweij: As discussed on the call yesterday, we want to have a subfolder per device pipeline - which we can mimic in the upload folder as well to make each pipeline upload their own .zip files. 

Which mean, download folder:
/device_type/patient_id/device_id
and upload folder:
/device_type/.zipfiles

I've adjusted the code to ensure the use of subfolders per device_type

data_transfer/jobs/shared.py
davidverweij: Woops - I think I meant `rmdir()` throws if not empty. I've added a catch now, but have not yet implemented a solution what to do (what checks to make if not empty)

davidverweij:
@jawrainey, I've amended the requested changes - please see latest commit

Issue #31, opened 2021-02-03
Integrating logging throughout each pipeline


jawrainey:

davidverweij:
Given our current logging implementation and as we slowly add logging throughout - I'd argue we can close this issue; @jawrainey ?

PR #44, opened 2021-03-05
 Integrate DMP upload from dmpy

### LINKED ISSUE #30: Integrate DMP upload
The [DMPy](https://github.com/ideafast/dmpy/compare/ft_upload) package has a branch that faciliates uploading data, but is not yet finalised while we wait on the new DMP release that will provide an `API_TOKEN`. Once this is finalised, we will need to do the following:

- [ ] Implement and merge the upload feature branch into master in [DMPy](https://github.com/ideafast/dmpy/compare/ft_upload).
- [ ] Add this git repo as a dependency in poetry within this middleware.
- [ ] Implement the [upload](https://github.com/ideafast/middleware-services/blob/a6ba0ad3e9040a0c6efd78ca4c2caf4adcf28264/data_transfer/services/dmpy.py#L16-L17) method within the dmpy service. This only needs done once since it's a shared task.
- [ ] Add associated tests

jawrainey:
## What Changed

Use the modified `dmpy` client to upload a file to the DMP. I've also added logging to show logs output from`dmpy.client` and that we can use elsewhere (i.e., #31). This resolves #30.

## Testing

1. Create a local file named `dmpy.env` and add the relevant credentials.
2. Download `51eb52d7dd8d5c6570e123f48b3d05898940aaf744ee68cfa9fdd1338237dfa2` file from the 'hackathon_test' dataset (ideally the dreem file that is 2.2GB) and store it in `/data/uploading`
3. Rename this file to `NXQEXRT-DRMHK9K4K-20210121-20210125` and make sure there are no other files in the upload folder.
4. As we only want to test the upload feature, comment all other pipeline stages in  `main.py`, i.e., [here](https://github.com/ideafast/middleware-services/blob/master/data_transfer/main.py#L45-L57)
5. Also comment out all lines in [`batch_upload_data` besides the upload](https://github.com/ideafast/middleware-services/blob/master/data_transfer/jobs/shared.py#L17-L28).
5. Run `python data_transfer/main.py` and observe the log output. Upload make take some time.
6. Once uploaded, verify on the DMP that the file was uploaded. 

Note: please review this in tandem with [upload implementation in dmpy](https://github.com/ideafast/dmpy/pull/2).


pyproject.toml
```diff
@@ -11,9 +11,10 @@ python = "^3.8"
 fastapi = "^0.61.1"
 uvicorn = "^0.12.2"
 requests = "^2.24.0"
-python-dotenv = "^0.14.0"
+python-dotenv = "^0.15.0"
 pymongo = "^3.11.2"
 boto3 = "^1.16.61"
+dmpy = { git = "https://github.com/ideafast/dmpy.git", branch = "ft_upload" }
```
jawrainey, line 17: Note: this branch should be changed to "master" (or removed) once [this issue](https://github.com/ideafast/dmpy/pull/2) is merged.

logging.conf
```diff
@@ -0,0 +1,40 @@
...
+level=DEBUG
+handlers=console,tofile
+qualname=dmpy.client
+propagate=0
+
+[handler_tofile]
+class=FileHandler
+level=DEBUG
+formatter=generic
+args=("./dtransfer.log",)
```
jawrainey, line 30: This was primarily for testing -- @davidverweij do we want to output this this to a file? Note line 22, so only the dmpy output is output into this file and also the console. I'd assume for now we only want console logs as they would work best in docker, etc

pyproject.toml
davidverweij: Agreed, I would argue to tag a version to the branch at that stage to allow developments on the master branch without breaking it's implementation here. E.g.,
```python
dmpy = { git = "https://github.com/ideafast/dmpy.git", tag = "v1.0" }
```

data_transfer/services/dmpy.py
```diff
@@ -1,6 +1,15 @@
...
 import shutil
 from pathlib import Path
 
+from dmpy.client import Dmpy
+from dmpy.core.payloads import FileUploadPayload
+
+from data_transfer.config import config
+from data_transfer.utils import wear_time_in_ms
+
+log = logging.getLogger(__name__)
```
davidverweij, line 11: Nice - this easily separates the logging per module

logging.conf
davidverweij: Yes, I'd say only console for now - and depending on how we are going to schedule the `data_transfer` runs, we could include a file (i.e. if we use a cronjob)

davidverweij (approved):
Nice one @jawrainey  :rocket:

I was a bit confused re: environmental variables, as we take a different approach to the DMPY and data_transfer, but it now makes sense due to dmpy operating as a import module. 

Minor: you forgot to mention I have to run `poetry install` 😄 ; and I'd suggest to add the `dmp_study_id` to the example .env file. Other than that - good to go. Happy for you to merge when amended.

pyproject.toml
davidverweij: Can we merge in this change after merging the PR on the dmpy side? Then we have that in one go.

data_transfer/config.py
```diff
@@ -15,6 +15,8 @@
 class Settings(BaseSettings):
     database_uri: str
 
+    dmp_study_id: str
+
```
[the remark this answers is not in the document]
jawrainey: My thinking here was that the user of the dmpy client might want to create multiple objects that upload to multiple study sites, i.e., to allow flexability. However, this would require creating multiple `config.py` since one maps to a study ID.

I could instead move this to `dmpy.env` as it is currently only set once?

logging.conf
jawrainey: Worth noting that console output will be picked up in docker contains for easy debugging, which we can pipe into a log outside of docker, etc

data_transfer/config.py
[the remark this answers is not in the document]
jawrainey: Note: [moved this to dmpy.env](https://github.com/ideafast/dmpy/pull/2/commits/25f2432117348034147ab840a51b4d13035ddde5) and [removed it from this repo](https://github.com/ideafast/middleware-services/pull/44/commits/24c84318ac0d55d340ada86f47bc9b1b473a4f42)

Issue #16, opened 2021-01-28
[Dreem]: verbose error handling and unit tests


jawrainey:
Current implementation was merged to progress other devices concurrently. Dreem implementation lacks in-depth error handling and unit test.

Issue #29, opened 2021-02-03
Automate data pipeline


jawrainey:
Currently we are implementing data pipelines that run locally while external services are being setup (e.g., UCAM), but once all services are in place and more device pipeline implementations we will automate each pipeline. Previously, we had designed an AMQP task queue, but looking into alternatives [apache airflow](https://airflow.apache.org/) seems ideal and so our current implementations are geared towards to use DAGs to make adoption of airflow easier.

Issue #24, opened 2021-02-02
Abstract `download_file` protocols across devices due to similarity


davidverweij:
Just a note for now, but this is very similar to the dreem implementation (which is fine), so perhaps in a future PR we can abstract it out into a shared task rather than device specific? since the only difference between dreem/VTT at the moment is how data is downloaded (line 103).

This is food for thought on how we might abstract this further once we implement the DAGs

_Originally posted by @jawrainey in https://github.com/ideafast/middleware-services/pull/21#discussion_r568535114_

davidverweij:
This now also includes other protocols, such as the recently introduced `uid_to_hash()` in combination with `all_hashes()` methods.

jawrainey:
Agreed -- also pulling out device agnostic logic such as `__device_id_from_ucam` in `dreem.py` since it determines a device in UCAM by a wear period and DeviceType (needs renaming too).

jawrainey:
Worth noting now that while there is some overlap, BTF differs in that it returns a filename rather than bool. **I argue that DRM and other pipelines should change to follow BTF for consistency.**

PR #46, opened 2021-03-19
Dockerise data transfer pipeline

jawrainey:
## What Changed?

1. Added dockerfile for datatransfer middleware
2. Modified docker-compose file to run all services for pipeline
3. Refactored `main.py` by pulling out `dags` to their own folder. This is helpful when running batch jobs
4. Refactored dreem DAG (`drm.py`) to operate on a per-patient level, which reduces the quantity of data stored on the server at once.
5. Running dreem pipeline as batchjob that invokes a DAG via the docker file. This is a temporary solution as we do not know which devices will make it into COS, so no need to fancy airflow (or celery) solution yet. 
6. Added solution to use development environmental variables when they are provided, e.g., for local development.
7. Refactored how `-meta.json` is stored to share logging/logic between DAGs.
8. Added error handling and logging to dreem library. 

Notes:

- `Dtransfer.Dockerfile` produces a large image (346MB) I can optimise this to ~100MB if necessary, but results in longer build times.
- We should pull out error handling a level higher, but imho should do this in another PR.
- I've not bothered updating `cli.py` yet to invoke docker build as we should pull out datatransfer to a new repo before then.
- Currently runs every minute: we probably should modify `Dtransfer.Dockerfile` so that setting when the pipeline is run is configurable, e.g., nightly.
 
## How to Test?

0. Modify code: Add `all_records = all_records[-9::3]` after the `dreem_api` call so that the pipeline operates on 3 files.
1. Build images: from the root folder `docker build -t "middleware" . ` to build the middleware image (consumer) and `docker build -f "Dtransfer.Dockerfile" -t "dtransfer" .` to build datatransfer image.
2. Run Services: run `docker-compose up -d`
3. Observe logs: run `docker logs -f dtransfer` to see the logs and note that it runs every minute (current config).
4. Check DMP: check hackathon dataset on DMP to see data uploaded

Closes: #16 and #29 (partially)


data_transfer/devices/dreem.py
```diff
@@ -108,33 +112,34 @@ def download_metadata(self) -> None:
...
-                    print(f"Metadata cannot be determined for:\n    {recording}")
+                    log.debug(
+                        f"Device does not exists in UCAM for {patient_id}:\n {recording}"
+                    )
+                    log.debug("Attempting to use inventory.")
                     continue
             else:
                 unknown += 1
-                print(f"Metadata cannot be determined for:\n    {recording}")
+                log.debug(f"Patient ID ({patient_id}) is not in UCAM:\n    {recording}")
```
davidverweij, line 122: Could you please implement the ID formatter and validator - as [in the ByteFlies branch?](https://github.com/ideafast/middleware-services/blob/3e65fca9a451f1fb4599b795e823709159683a77/data_transfer/devices/byteflies.py#L113-L115). This also shows that we keep the `-` throughout the pipeline and only remove it when translating to uploads to the DMP.

davidverweij (commented):
Have to jump, but will continue review later. See initial comments if you want something to work on :)

data_transfer/devices/dreem.py
```diff
@@ -88,53 +95,37 @@ def download_metadata(self) -> None:
...
+                log.debug(f"Patient ({patient_id}) or Device ID ({device_id}) missing.")
                 unknown += 1
-                print(f"Metadata cannot be determined for:\n    {recording}")
-                continue
+                continue  # Skip record
+            known += 1
 
             record = Record(
-                filename=recording.id,
+                hash=recording.id,
```
davidverweij, line 110: Can you please use the [hasher](https://github.com/ideafast/middleware-services/blob/17a9674bf803deb619a5e507a11549af72cc8706/data_transfer/utils/__init__.py#L151-L157) to set this value? 

In the BTF branch, I calculate this hash for each recording and add it to the metadata. I'd suggest to do the same for this DRM branch. Effectively, in the DRM lib before [this line](https://github.com/ideafast/middleware-services/blob/17a9674bf803deb619a5e507a11549af72cc8706/data_transfer/lib/dreem.py#L57), add :
```python
result["IDEAFAST"]["hash"] = uid_to_hash( result["id"], DeviceType.DRM )
```

Then this line here, should become:
```python
hash=item["IDEAFAST"]["hash"]
```
which is identical to the BTF implementation

data_transfer/devices/dreem.py
```diff
@@ -59,9 +63,13 @@ def download_metadata(self) -> None:
         # Note: includes metadata for ALL data records, therefore we must filter them
         all_records = dreem_api.get_restricted_list(self.session, self.user_id)
 
+        log.info(f"Total dreem records: {len(all_records)} for {self.study_site}")
+
         # Only add records that are not known in the DB based on stored filename
         # i.e. (ID and filename in dreem)
         unknown_records = [r for r in all_records if r["id"] not in set(all_hashes())]
```
davidverweij, line 70: See comment at 110 below, but when using the hash, `r["id"]` becomes `r["IDEAFAST"]["hash"]`

data_transfer/devices/dreem.py
```diff
@@ -88,53 +95,37 @@ def download_metadata(self) -> None:
...
+                end_wear=recording.end,
             )
 
             create_record(record)
-            print(f"Record Created:\n   {record}")
 
-            path = Path(config.storage_vol / f"{record.filename}-meta.json")
             # Store metadata from memory to file
-            utils.write_json(path, item)
+            utils.write_json(record.metadata_path(), item)
```
davidverweij, line 122: When you have the added hash, and following the BTF implementation, this line should be preceded by:
```python
del item["IDEAFAST"]  # remove record-specific temporary metadata
```

data_transfer/devices/dreem.py
```diff
@@ -145,27 +136,49 @@ def __recording_metadata(self, recording: dict) -> DreemRecording:
...
+                device = ucam.record_by_wear_period_in_list(devices, start, end)
+                return device.device_id if device else None
+        return None
+
     def __patient_id_from_ucam(
-        self, device_serial: str, dreem_start: datetime, dreem_end: datetime
+        self, device_serial: str, start: datetime, end: datetime
     ) -> Optional[str]:
         """
         Determine PatientID by wear period of device in UCAM.
```
davidverweij, line 165: Could you add to this doc comment that it is relying on the inventory as well?

data_transfer/lib/dreem.py
```diff
@@ -58,39 +50,44 @@ def get_restricted_list(session: requests.Session, user_id: str) -> List[dict]:
...
-        url = result["next"]
-        results.extend(result["results"])
+        try:
+            response = session.get(url)
+            response.raise_for_status()
+            result: dict = response.json()
+            url = result["next"]
+            results.extend(result["results"])
+        except requests.HTTPError:
+            # attempt to skip this one record
```
davidverweij, line 60: Did you get this error before? Just curious - but will this ever jump out of the while loop if it keeps throwing? (As it potentially reuses the same `result["next"]`) Perhaps narrow the error down to the specific HTTPError?

data_transfer/tasks/shared.py
```diff
@@ -0,0 +1,46 @@
...
+
+    for extension in [FILE_TYPES[device_type.name], "-meta.json"]:
+        fname = f"{record.filename}{extension}"
+
+        old_path = data_input / fname
+        new_path = destination / fname
+
+        old_path.rename(new_path)
+        record.is_prepared = True
+        update_record(record)
```
davidverweij, line 46: Why is this whole file here? I see only SMA refers to this method, but this has been replaced some time ago by `jobs/shared.py` --> prepare_data_folders. This file can be deleted.

davidverweij (changes requested):
Pfew, quite an extensive review this one! I dropped quite some comments here and there, so please respond and we can discuss my suggestions.

Overall looking good and tests went well. Once this is merged I can translate the changes back into the BTF branch and start working on the VTT end this week.

.dtransfer.prod.env.example
```diff
@@ -1,16 +1,14 @@
-# IDEAFAST
-INVENTORY_API="http://0.0.0.0:8000/inventory/"
-DATABASE_URI="mongodb://localhost:27017/"
+# Switch between Dev/Prod
+IS_DEV="True"
+
+# IDEAFAST (WP3)
+INVENTORY_API=""http://host.docker.internal:80/inventory/""
+# Change USER, PASSWORD, and DATABASE accordingly
+DATABASE_URI="mongodb://USER:PASSWORD@mongo:27017/?authSource=DATABASE&authMechanism=SCRAM-SHA-256"
```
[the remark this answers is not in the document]
jawrainey: Yes, for prod development it should be the host name as it maps to an IP to communicate across containers 👍🏼

data_transfer/main.py
```diff
@@ -1,99 +1,25 @@
...
-
-    # Data is finalised and moved to a folder in /uploading/
-    shared_jobs.prepare_data_folders(DeviceType.BTF)
-    # All said folders FOR ALL DEVICES are uploaded once per day
-    shared_jobs.batch_upload_data(DeviceType.BTF)
+from data_transfer.config import config
+from data_transfer.dags import btf, drm, sma
+from data_transfer.utils import DeviceType, StudySite
 
+fileConfig("logging.ini", disable_existing_loggers=False)
```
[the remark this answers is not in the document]
jawrainey: Default is True, so can remove that parameter. It would be possible to have two logging streams and for to be output to a file. Perhaps something we can look into in a seperate PR?

README.md
```diff
@@ -35,39 +35,44 @@ When developing the consumer API run:
...
     poetry run compose
 
-We use [Docker Hub](https://hub.docker.com/u/ideafast) to store images. To push
-to your own image, update `REGISTRY` inside `cli.py` and run:
+*Note:* for local development docker is primarily used to test changes prior to deploying live _or_ to interact with services in isolation.
+
+### Deploying
+
+[Docker Hub](https://hub.docker.com/u/ideafast) is used to store images. To push
+to a `$REPO` run  the following:
```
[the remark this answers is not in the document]
jawrainey: Repo is a repoistory similar to GitHub that stores a list of images, e.g., `ideafast/middleware` or `mongo`. Alternatively, we could name this `IMAGE_NAME`. We would want the image name to be a repo for docker push to work without any changes

data_transfer/config.py
```diff
@@ -2,85 +2,111 @@
...
 
-# TODO: would be docker volume
-root_path = Path(__file__).parent.parent
 
+class GlobalConfig(BaseSettings):
+    """Global configurations."""
 
-class Settings(BaseSettings):
-    database_uri: str
+    root_path = Path(__file__).parent.parent
```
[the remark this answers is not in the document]
jawrainey: Yeah, unless dtransfer was in `/src/dtransfer` or such

data_transfer/config.py
```diff
@@ -2,85 +2,111 @@
...
+
+    storage_vol: Path = data_path / "input"
+    upload_folder: Path = data_path / "uploading"
+
+    byteflies_devices = root_path / "byteflies_devices.csv"
 
-    # UCAM
-    ucam_data: Path = root_path / "ucam_db.csv"
+    dreem_users: Path = csvs_path / "dreem_users.csv"
+    dreem_devices: Path = csvs_path / "dreem_devices.csv"
```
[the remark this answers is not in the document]
jawrainey: We could alternatively at the folder `/local` via `.gitkeep` and add a README in there of expected files?

data_transfer/devices/dreem.py
jawrainey: How come we add it to the metadata? I could certainly change this to use `uid_to_hash` and update the set difference above, but it's unclear why we would send this middleware-specific hash to the DMP?

data_transfer/lib/dreem.py
jawrainey: I received one of the derived errors (`RequestException` I think). `HTTPError` is similar to `Exception` in that it contains all requests specific potential errors. Good point -- the `result[next]` will be different in each iteration so won't be infinite-lopped, so in that case it'll catch as many errors as the length of results.

data_transfer/devices/dreem.py
```diff
@@ -73,11 +81,10 @@ def download_metadata(self) -> None:
...
             # Serial may not exist in lookup, e.g., if Dreem send a device replacement.
             if not device_serial:
-                print(f"Unknown Device:\n   {recording}")
-                # Move onto next record: skips logic below to simplify error handling
-                continue
+                log.debug(f"Unknown Device:\n   {recording}")
+                continue  # Skip record
 
-            patient_id = (
+            patient_id = utils.format_id_patient(
```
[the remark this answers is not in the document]
jawrainey: Do we need to know this, and could that not be something that is logged within `format_id_patient` so that it can be shared across device implementations?

In the case of dreem, an error is logged if DeviceID or PatientID is None.

data_transfer/devices/dreem.py
davidverweij: @jawrainey , This removes all metadata generated temporarily for the pipeline

data_transfer/main.py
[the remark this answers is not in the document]
davidverweij: Separate PR - agreed!

data_transfer/config.py
```diff
@@ -2,85 +2,111 @@
...
     byteflies_password: str
     byteflies_aws_client_id: str
     byteflies_aws_auth_url: str
 
-    # NOTE: Is this the best place to store this?...
     byteflies_group_ids: dict = {
         StudySite.Kiel: "ba92eb10-74d8-11ea-9162-a946985733fd",
         StudySite.Rotterdam: "8c5ed850-b789-11ea-870a-c3400b84381c",
         StudySite.Newcastle: "2f5e6630-4c03-11ea-9f2d-0949ce20b25c",
         StudySite.Muenster: "aa3f7660-ba2a-11ea-bb28-b3fd87020c94",
```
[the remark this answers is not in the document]
jawrainey: Best to load em' from .env, but now they're public we might as well leave them?

data_transfer/devices/dreem.py
jawrainey: Resolved via adding `__unknown_records` method

data_transfer/devices/dreem.py
```diff
@@ -224,6 +225,12 @@ def download_file(self, mongo_id: str) -> None:
             self.session, record.download_folder(), record.manufacturer_ref
         )
 
+        # Useful metadata for performing pre-processing.
+        downloaded_file = Path(
+            record.download_folder() / f"{record.manufacturer_ref}.h5"
+        )
+        record.meta = dict(filesize=downloaded_file.stat().st_size)
```
davidverweij, line 232: This would override existing metadata I believe - perhaps directly set the field? i.e.:
```python
record.meta.filesize = downloaded_file.stat().st_size

```

data_transfer/devices/dreem.py
```diff
@@ -229,7 +229,7 @@ def download_file(self, mongo_id: str) -> None:
         downloaded_file = Path(
             record.download_folder() / f"{record.manufacturer_ref}.h5"
         )
-        record.meta = dict(filesize=downloaded_file.stat().st_size)
+        record.meta["filesize"] = downloaded_file.stat().st_size
```
davidverweij, line 232: I'm not there with testing (sorry, taking a long time to create all records), but given that `meta` is optional, it will be `None` by default. Then, accessing `["filesize"]` might throw an error. I'll test in a bit, but if so - we need to update the `Record`:

```python
class Record(BaseModel):
    ...

    # Relevant to individual device pipelines
    meta: {}      # default. Indicates type and allows to be empty

    ...
```

data_transfer/dags/drm.py
```diff
@@ -0,0 +1,46 @@
...
+            mongoid = dreem_tasks.task_download_data(dreem, record.id)
+            dreem_tasks.task_preprocess_data(mongoid)
+        # Only upload when all records are ready
+        if all_records_downloaded(records):
+            log.debug(f"All records for {patient_device} DOWNLOADED -> PREPARING ...")
+            shared_jobs.prepare_data_folders(DeviceType.DRM)
+            log.debug(f"All records for {patient_device} PREPARED   -> UPLOADING ...")
+            shared_jobs.batch_upload_data(DeviceType.DRM)
+        else:
+            log.error(f"Some records for {patient_device} were not downloaded.")
```
davidverweij, line 46: Excellent solution - makes a lot of sense to detect any non-uploaded, and then try to rerun any part of the pipeline that is still needed.

data_transfer/devices/dreem.py
```diff
@@ -88,53 +96,56 @@ def download_metadata(self) -> None:
...
-                    # Determine usage based on weartime as it exists in UCAM
-                    device_record = ucam.record_by_wear_period_in_list(
-                        dreem_devices, recording.start, recording.end
-                    )
-                # Edge-case: device not logged with patient in UCAM
-                else:
-                    print(f"Metadata cannot be determined for:\n    {recording}")
-                    continue
-            else:
+            if not (patient_id := utils.format_id_patient(_patient_id)):
```
davidverweij, line 99: I think this is good to understand where errors come from. Although isolated, this will allow us to detect possible formatting errors; even though our validator will soon allow a bit more lossy written ids.

data_transfer/devices/dreem.py
```diff
@@ -88,53 +96,56 @@ def download_metadata(self) -> None:
...
                 unknown += 1
-                print(f"Metadata cannot be determined for:\n    {recording}")
                 continue
 
+            device_id = self.__device_id_from_ucam(
+                patient_id, recording.start, recording.end
+            ) or inventory.device_id_by_serial(device_serial)
+
+            if not patient_id or not device_id:
+                log.error(f"Metadata cannot be determined for {recording}")
```
davidverweij, line 111: :rocket:

data_transfer/devices/dreem.py
```diff
@@ -88,53 +96,56 @@ def download_metadata(self) -> None:
...
+        """
+        Only add records that are not known in the DB, i.e., ID and filename.
+        """
+        results = {}
+        known_records = all_hashes()
+        for record in records:
+            record_hash = uid_to_hash(record["id"], self.device_type)
+            if record_hash not in known_records:
+                results[record_hash] = record
+        return results
```
davidverweij, line 144: I think it is a fair solution to write this method per-device pipeline as they focus on different keys in the `item` object. We can look at pulling this out along with #24 to ensure the result is equal across pipelines.

data_transfer/devices/dreem.py
davidverweij: :rocket:

data_transfer/schemas/record.py
```diff
@@ -21,7 +24,10 @@ class Record(BaseModel):
     patient_id: str
     start_wear: datetime
     end_wear: datetime
-    meta: dict  # meta data relevant to individual device pipelines
+
+    # Relevant to individual device pipelines
+    meta: dict = {}
```
davidverweij, line 29: 👍

data_transfer/tasks/dreem.py
```diff
@@ -15,6 +15,7 @@ def task_preprocess_data(mongoid: str) -> str:
     Preprocessing tasks on dreem data.
     """
     record = read_record(mongoid)
-    record.is_processed = True
-    update_record(record)
+    if not record.is_processed:
+        record.is_processed = True
+        update_record(record)
```
davidverweij, line 20: Much processing, such wow!

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "consumer"
 version = "0.1.0"
 description = "Hosts the middleware services for IDEA-FAST, including: data transfer protocols and application consumer APIs."
-authors = ["Drew Miley <drew.miley@ncl.ac.uk>", "Jay Rainey <jay.rainey@ncl.ac.uk>"]
+authors = ["Jay Rainey <jay.rainey@ncl.ac.uk>", "David Verweij <david.verweij@ncl.ac.uk>"]
```
davidverweij, line 5: 😊

pyproject.toml
```diff
@@ -14,7 +14,8 @@ requests = "^2.24.0"
 python-dotenv = "^0.15.0"
 pymongo = "^3.11.2"
 boto3 = "^1.16.61"
-dmpy = {git = "https://github.com/ideafast/dmpy", rev = "0.1.1"}
+mypy-boto3-s3 = "^1.17.22"
```
davidverweij, line 17: @jawrainey Don't this need to be a `dev` dependency? As it is only used for type checking?

davidverweij (approved):
All good! Thanks for all the requested changes - just making sure it makes sense now and in the future 🚀

pyproject.toml
davidverweij: NVM - as you say, it's needed since we declare types with this package.

PR #51, opened 2021-03-25
Refer to logging configuration via config path

jawrainey:
The location of the logging fle is hardcoded and can therefore be invalid depending on where the python script is run from. Instead, I have moved this path into config and used the root_path as needed.  This resolves a 🐛 in current master. To reproduce:

1. Build docker image locally and run pipeline -- note it fails
2. Switch to this branch and run pipeline -- note success.


davidverweij (approved):
Not much to say! :rocket:

Issue #28, opened 2021-02-03
Mock download in data pipeline when running locally


jawrainey:
When running pipelines locally we do not want to download files from external manufacters primarily due to file sizes. As such, we could have something similar to the following:

Add an environmental variable `IS_LOCAL=True` to our `.env` configuration, then where we call `download_file` in each pipeline we could decorator it. The decoractor could then not call the download_data method if the env variable is set:

```python
def skip_if_is_local(download_data):
 def wrapper(session, record_id):
   if config.is_local:
     file_path = Path(config.storage_vol) / f"{​record_id}​.h5" # NOTE: hard-coded dreem extension
     write_json(file_path, {​}​)
     # NOTE: we do not run the download method below ;-)
   else:
     download_data(session, record_id)
 return wrapper

@skip_if_is_local
def download_file(session: requests.Session, record_id: str) -> bool:
   ....
```

This above implementation is for dreem, but of course the parameters for `download_file` may vary, e.g., such as for [VTT SMA](https://github.com/ideafast/middleware-services/blob/master/data_transfer/devices/vttsma.py#L103): 

```python
def download_files(bucket: Bucket, patient_hash: str, export_date: str,) -> bool: 
   ...
```
Related to this, you will see above that I have hard-coded the extension. This is quite an important detail as the shared task expected specific file types: 

https://github.com/ideafast/middleware-services/blob/a6ba0ad3e9040a0c6efd78ca4c2caf4adcf28264/data_transfer/tasks/shared.py#L6-L10 

Therefore:

1. Decorator above should accept filetype as a parameter
2. Accept arbitary arguments (**args) where the first one must be the filename or keyword argument (**kwargs) such that we can specific `kwargs['record_id']`

This might result in something like the following (not tested):

```python
def skip_if_is_local(filetype):
    def decorator(download_data):
        def wrapper(*args):
            if config.is_testing:
                file_path = Path(config.storage_vol) / f"{args[0]}{filetype}"
                write_json(file_path, {})
            else: 
               download_data(session, record_id)
        return wrapper
    return decorator

from data_transfer.tasks.shared import FILE_TYPES

@skip_if_is_local(FILE_TYPES['DRM'])
def download_file(session: requests.Session, record_id: str) -> bool:
   ....
```

jawrainey:
Simple solution to the above might be to store the file extension within the `filename` in `Record`? Since  there is clearly a minor dependency between data download and preparation when there shouldn't be

[...]

PR #55, opened 2021-03-29
Byteflies update, closes #27

### LINKED ISSUE #27: Initial Data protocol for Byteflies
Similar to Dreem (#14) and VTT (#21), the data transfer protocol for ByteFlies needs to be developed. Initial work has been done in [the byteflies](https://github.com/ideafast/middleware-services/compare/byteflies) branch.

Initial tasks to get this up and running include:

- [x] branch into byteflies2
- [x] mimick the protocol for ByteFlies
- [x] setup a ByteFlies DAG for potential use with Apache Airflow
- [ ]  adapt the 4 stages (pull meta, pull data, preprocess data, upload data)

Please comment/add any steps I am missing here. Will potentially set up a project with milestones for this issue.

Docs on https://ifu.byteflies.net/docs/dev/api-reference.html

davidverweij:
These changes ensure that the ByteFlies branch follows a similar flow to the Dreem branch, including folder structure, batch-processing by patient+device_id as well as logging. 

Partially responds to #28 (at least for testing) 

## Test
- pull down branch
- run `poetry install`
- run `poetry shell`
- run `poetry run nox -rs` for all tests
- run `poetry run consumer` to run the inventory for the overall pipeline
- ensure that you have the mongo DB running
- ensure your `.envs` are correctly set - including `ACCESS_AUTHENTICATED_ENDPOINTS=True` in `.consumer.env`
- run `python data_transfer/main.py BTF Kiel 0 1` to run the BTF pipeline for today (`0`) and yesterday (`1`)

For testing the historical data pipeline, I suggest to add `return` [after logging the data period here](https://github.com/ideafast/middleware-services/blob/352459b483275c166011d71ff435bf17a1aeb2bb/data_transfer/dags/btf.py#L49), and run:
- `python data_transfer/main.py BTF Kiel -1`

### DIFF
--- a/data_transfer/dags/btf.py
+++ b/data_transfer/dags/btf.py
@@ -65,5 +65,3 @@ def dag(
             shared_jobs.batch_upload_data(DeviceType.BTF)
         else:
             log.error(f"Some records for {patient_device} were not downloaded.")
-
-        break
--- a/data_transfer/dags/btf.py
+++ b/data_transfer/dags/btf.py
@@ -68,13 +68,15 @@ def dag(
             log.error(f"Some records for {patient_device} were not downloaded.")
 
 
-def historical_dag(study_site: StudySite, delta: int) -> None:
+def historical_dag(
+    study_site: StudySite, _days: Union[int, str] = -1, delta: Union[int, str] = 0
+) -> None:
     """Loops through set periods to retreive all historical data"""
     days_to_cover = (
         datetime.today() - datetime.fromisoformat(config.byteflies_historical_start)
     ).days
     # jump straight to delta reference and skip days in between
-    tracker = delta
+    tracker = int(delta)
     while tracker < days_to_cover:
         dag(study_site, 50, tracker)
         # ensure 1 day overlap for sanity
--- a/data_transfer/dags/btf.py
+++ b/data_transfer/dags/btf.py
@@ -1,13 +1,17 @@
+import logging
 from datetime import datetime
 
-from data_transfer.db import records_not_uploaded
+from data_transfer.db import all_records_downloaded, records_not_uploaded
+from data_transfer.devices.byteflies import Byteflies
 from data_transfer.jobs import byteflies as byteflies_jobs
 from data_transfer.jobs import shared as shared_jobs
 from data_transfer.tasks import byteflies as byteflies_tasks
 from data_transfer.utils import DeviceType, StudySite, get_period_by_days
 
+log = logging.getLogger(__name__)
 
-def dag(study_site: StudySite) -> None:
+
+def dag(study_site: StudySite, timespan: int, now: int = -1) -> None:
     """
     Directed acyclic graph (DAG) representing dreem data pipeline:
 
@@ -17,24 +21,40 @@ def dag(study_site: StudySite) -> None:
             ->task_prepare_data
         ->batch_upload_data
 
+    Parameters
+    ----------
+    study_site : StudySite
+    timespan : int
+        The amount of days in the past to query data for. This should
+        correspond with the pipeline recurrance + margin to ensure overlap
+    now : int, optional
+        The unix time to reference timespan to, defaults to today for the pipeline.
+        Allows traversing and stepping through the past for historical data
+
+
     NOTE/TODO: this method simulates the pipeline.
     """
 
-    # TODO: ensure that we get the End of Day as intended based on automation schedule
-    # TODO: should we ensure overlap with previous run? How often do we run this DAG?
-    # TODO: perhaps pull this one abstraction higher (into the init of this DAG?)
-    data_period = get_period_by_days(datetime.today(), 2)  # NOTE: two days for testing
+    byteflies = Byteflies()
+
+    reference: datetime = datetime.today() if now == -1 else datetime.fromtimestamp(now)
+    data_period = get_period_by_days(reference, timespan)
     byteflies_jobs.batch_metadata(*data_period, study_site)
 
     results = records_not_uploaded(DeviceType.BTF)
 
-    for _patient_device, records in results.items():
+    # NOTE: group records by patients per device to process small batches.
+    for patient_device, records in results.items():
         for record in records:
             # Each task should be idempotent. Returned values feeds subsequent task
-            mongoid = byteflies_tasks.task_download_data(record.id)
+            mongoid = byteflies_tasks.task_download_data(byteflies, record.id)
             byteflies_tasks.task_preprocess_data(mongoid)
 
-        # Data is finalised and moved to a folder in /uploading/
-        shared_jobs.prepare_data_folders(DeviceType.BTF)
-        # All said folders FOR ALL DEVICES are uploaded once per day
-        shared_jobs.batch_upload_data(DeviceType.BTF)
+        # Only upload when all records are ready
+        if all_records_downloaded(records):
+            log.debug(f"All records for {patient_device} DOWNLOADED -> PREPARING ...")
+            shared_jobs.prepare_data_folders(DeviceType.BTF)
+            log.debug(f"All records for {patient_device} PREPARED   -> UPLOADING ...")
+            shared_jobs.batch_upload_data(DeviceType.BTF)
+        else:
+            log.error(f"Some records for {patient_device} were not downloaded.")
--- a/data_transfer/dags/btf.py
+++ b/data_transfer/dags/btf.py
@@ -2,6 +2,7 @@ import logging
 from datetime import datetime
 from typing import Union
 
+from data_transfer.config import config
 from data_transfer.db import all_records_downloaded, records_not_uploaded
 from data_transfer.devices.byteflies import Byteflies
 from data_transfer.jobs import byteflies as byteflies_jobs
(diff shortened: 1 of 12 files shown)


data_transfer/dags/btf.py
```diff
@@ -1,13 +1,21 @@
...
 from data_transfer.tasks import byteflies as byteflies_tasks
 from data_transfer.utils import DeviceType, StudySite, get_period_by_days
 
+log = logging.getLogger(__name__)
 
-def dag(study_site: StudySite) -> None:
+
+def dag(
+    study_site: StudySite, days: Union[int, str] = 50, delta: Union[int, str] = 0
+) -> None:
```
jawrainey, line 18: I'd make the parameters explict types (ints) as these are defined once in the CLI (main.py) and can be casted there once rather than below and elsewhere. The docstring also indicates these are ints, but that they are optional which isn't quite correct?

data_transfer/dags/btf.py
```diff
@@ -17,24 +25,59 @@ def dag(study_site: StudySite) -> None:
...
+    delta : int, optional
+        Initial reference (0 == today) to query backwards from. This allows
+        traversing back in time for historical data
+
+
     NOTE/TODO: this method simulates the pipeline.
     """
+    byteflies = Byteflies(study_site)
+
+    data_period = get_period_by_days(int(delta), int(days))
```
jawrainey, line 43: How come `get_period_by_days` accepts ints and returns strings, both of which are casted to and from?

I'd change `get_period_by_days` to accept and returns ints, then modify `batch_metadata` so the cast occurs once.

data_transfer/dags/btf.py
```diff
@@ -17,24 +25,59 @@ def dag(study_site: StudySite) -> None:
...
+            shared_jobs.prepare_data_folders(DeviceType.BTF)
+            log.debug(f"All records for {patient_device} PREPARED   -> UPLOADING ...")
+            shared_jobs.batch_upload_data(DeviceType.BTF)
+        else:
+            log.error(f"Some records for {patient_device} were not downloaded.")
+
+
+def historical_dag(
+    study_site: StudySite, _days: Union[int, str] = -1, delta: Union[int, str] = 0
+) -> None:
```
jawrainey, line 73: Same as above -- I would be explict with use of ints

data_transfer/dags/btf.py
```diff
@@ -17,24 +25,59 @@ def dag(study_site: StudySite) -> None:
...
+    """Loops through set periods to retreive all historical data"""
+    days_to_cover = (
+        datetime.today() - datetime.fromisoformat(config.byteflies_historical_start)
+    ).days
+    # jump straight to delta reference and skip days in between
+    tracker = int(delta)
+    while tracker < days_to_cover:
```
jawrainey, line 80: How come `tracker` is defined? You could use `delta` in the while loop in the same way?

data_transfer/devices/byteflies.py
```diff
@@ -77,51 +83,72 @@ def download_metadata(
         """
         # Note: includes metadata for ALL data records, therefore we must filter them
         all_records = byteflies_api.get_list(
-            self.session, config.byteflies_group_ids[study_site], from_date, to_date
+            self.session,
+            config.byteflies_group_ids[self.study_site],
+            from_date,
+            to_date,
```
jawrainey, line 89: How come these types are strings and not ints?

data_transfer/lib/byteflies.py
```diff
@@ -96,32 +100,38 @@ def get(self, signal_id: str, algorithm_id: str = "") -> Any:
...
 def download_file(
-    download_folder: Path,
     session: requests.Session,
+    download_folder: Path,
     studysite_id: str,
     recording_id: str,
     signal_id: str,
     algorithm_id: str = "",
-) -> bool:
+) -> Union[bool, str]:
```
jawrainey, line 109: This varied return type seems unsual? Maybe `Optional[str]` would be more explict in the intended outcome, i.e., a filename?

data_transfer/lib/byteflies.py
```diff
@@ -96,32 +100,38 @@ def get(self, signal_id: str, algorithm_id: str = "") -> Any:
...
 
-    # TODO: for now, assumes that this method never throws ...
-    __download_file(download_folder, url, filename)
-    return True
+        if __download_file(download_folder, url, filename):
+            return filename
+        return False
+    except requests.HTTPError:
+        log.error(f"GET Exception to {url} ", exc_info=True)
+        return False
```
jawrainey, line 134: I would return `None` in place of `False` given it would have the same outcome in an IF statement and is more descriptive of intent, i.e., if there is no filename then the string is non-existing. We could use "" instead, but imho None is more explict.

data_transfer/main.py
```diff
@@ -22,4 +30,9 @@
     if device == DeviceType.SMA:
         sma.dag()
     if device == DeviceType.BTF:
-        btf.dag(study_site)
+        # if 'days' == -1, trigger full history
+        if len(sys.argv) >= 4 and int(sys.argv[3]) == -1:
+            btf.historical_dag(study_site, *sys.argv[3:])
+        else:
+            # passes timespan and reference if present
```
jawrainey, line 37: Before `if len(...)`  I would define the variables as arguments and cast them once as needed. This mirrors the potential use of `click` or other CLIs where arguments are explictely defined. This also means you can cast those once and define intended input in both `dag` and `historical_dag` rather than using `Union` or such

data_transfer/utils/__init__.py
```diff
@@ -47,10 +47,16 @@ def format_weartime_from_timestamp(period: int) -> datetime:
...
+
+    # ensure end at night, but start in the morning to capture everything _those days_
+    end_date = reference.replace(hour=23, minute=59, second=59, microsecond=999999)
+    from_date = end_date - timedelta(
+        days=days, hours=23, minutes=59, seconds=59, microseconds=999999
+    )
+
     begin = str(int(from_date.timestamp()))
     end = str(int(end_date.timestamp()))
     return (begin, end)
```
jawrainey, line 62: Unclear why these are casted to strings? I would return timestamps and cast only where needed, e.g., rather than [casting back to its original type](https://github.com/ideafast/middleware-services/blob/352459b483275c166011d71ff435bf17a1aeb2bb/data_transfer/dags/btf.py#L45-L49)

This would mean that some uses of it would need to be modified, e.g., [here](https://github.com/ideafast/middleware-services/blob/352459b483275c166011d71ff435bf17a1aeb2bb/data_transfer/jobs/byteflies.py#L4)

tests/data_transfer_tests/byteflies/conftest.py
```diff
@@ -70,10 +71,16 @@ def download_callback(
...
 
 @pytest.fixture
-def populated_db() -> Collection:
-    db = mongomock.MongoClient().db
+@patch.object(byteflies.Byteflies, "authenticate")
+@patch.object(byteflies, "StudySite")
+def populated_db(mock_authenticate: Mock, mock_city: Mock) -> Collection:
+    # path db with mock
+    mock_db = mongomock.MongoClient().db
+    # get mock data
```
jawrainey, line 79: 🚀

tests/data_transfer_tests/byteflies/test_byteflies.py
```diff
@@ -1,45 +1,47 @@
...
 from data_transfer.db import main as db
-from data_transfer.devices import byteflies as devices
 from data_transfer.lib import byteflies as lib
-from data_transfer.tasks import byteflies as tasks
-from data_transfer.utils import DeviceType
 
 
-def test_get_list(mock_requests_session: dict) -> None:
+@patch.object(lib.time, "sleep")
+def test_get_list(mock_time_sleep: Mock, mock_requests_session: dict) -> None:
```
jawrainey, line 12: Great use of mocks throughout your tests -- great work 🚀

data_transfer/dags/btf.py
davidverweij: I'm casting here because due to our CLI use of this library, `delta` and `days` might be string

data_transfer/dags/btf.py
davidverweij: I'm casting here because due to our CLI use of this library, `delta` and `days` might be string

data_transfer/dags/btf.py
davidverweij: Good point!

data_transfer/lib/byteflies.py
davidverweij: Happy to do that, was just following the Dreem implementation (which returns a `bool`)

data_transfer/main.py
davidverweij: Am fixing this as we speak - but note that using `None` does not allow picking up the default and will require additional logic in the methods itself. 

Also, leaving out `click` implementation due to priorities

davidverweij:
@jawrainey responded to your comments - please have another look.

jawrainey:
Great work @davidverweij. This works for me locally (also in Docker locally) to get all existing records for Kiel/Munster. Adding the option to get historical data works perfectly too. Super job on adding quite a range of complex testing.

I've two very minor comments, but otherwise it's functionally perfect:

1. As [`__key_by_value`](https://github.com/ideafast/middleware-services/blob/byteflies_update/data_transfer/lib/byteflies.py#L140-L148) is not used in `byteflies.py` we could remove it?
2. Are the print statements [here](https://github.com/ideafast/middleware-services/blob/10ea22f67dea062b55a9682531afa74576fcc220/data_transfer/devices/byteflies.py#L263-L274) intended to exist or should these be `log.debug` instead?

davidverweij:
@jawrainey Good catches. I've removed the `__key_by_value`. The print statements were a sloppy leftover of my testing. I've also re-added the time.sleep(4) as you have for DRM to mitigate the (current) concerns around overloading the inventory api.

PR #56, opened 2021-03-29
Inventory fix

davidverweij:
Fixed the inventory comparison issue due to the the missing microseconds in `normalise_day`. Also added three test (as a starter) to ensure the inventory code  behaves as expected - though the inventory itself is thereby not tested (yet).

Note that this branches off master, hence a test or two in the ByteFlies branch will not pass.

## Test
- run `poetry run nox -rs tests`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,5 +16,6 @@ ucam_db*.csv
 dmypy.json
 *.log
 *.json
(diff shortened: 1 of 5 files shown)


jawrainey:
LGTM

PR #59, opened 2021-03-30
remove logs for URL requests; overloading the log

davidverweij:
Teensy fix - the logs were overloaded with URL responses.


jawrainey:
@davidverweij : LGTM -- do we want to log `result.headers` (e.g., to know expected filesize, etc) or shall we skip it given the quantity of BTF records?

PR #63, opened 2021-04-01
Cache access to serial IDs to improve performance

jawrainey:
## Problem

The inventory has a rate limit of about 150 requests per min. As such, we use `time.sleep(4)` in the DRM/BTF pipelines. However, this means they can take a considerably long time to run. Discussed further [here](https://github.com/ideafast/stack/issues/9). This is primarily only when we need to determine Device ID by serial on the inventory.

## Solution

- The inventory has an endpoint that returns ALL device data per model, e.g., DRM/BTF and @davidverweij suggested we use that endpoint instead [see here](https://github.com/ideafast/middleware-services/pull/42#discussion_r577032306). 
- I have implemented this endpoint in the consumer app, then access it as a service in data_transfer. 
- Critically, I have cached the results of one API request to the consumer middleware.

## Testing

- Run `python data_transfer/main.py DRM kiel` from master and note how long it takes to obtain metadata.
- Run the same from this branch 🚀 
- Run the tests
- Review the code


data_transfer/services/inventory.py
```diff
@@ -9,27 +9,40 @@
...
 
 @lru_cache
 def record_by_device_id(
     device_id: str, start_wear: datetime, end_wear: datetime
 ) -> Optional[Any]:
-    device_wears = [i for i in device_history(device_id).values()]
+    history = device_history(device_id)
+    if not history:
+        return None
+    device_wears = [i for i in history.values()]
```
jawrainey, line 45: Might want to add additional tests for these (and rest of services.inventory) -- perhaps in another PR?

davidverweij: Why are you using a kist comprehension here? `values()` inherently returns a `list`.

tests/data_transfer_tests/services/test_inventory.py
```diff
@@ -36,3 +36,34 @@ def test_normalise_day() -> None:
...
+
+def test_all_devices_by_type_cache_success() -> None:
+    inventory.all_devices_by_type.cache_clear()
+    num_requests = 10
+
+    with patch("requests.get", return_value=Mock()) as _:  # act
+        for __ in range(0, num_requests):
+            inventory.all_devices_by_type(utils.DeviceType.BTF)
+
+        hits, misses, _, __ = inventory.all_devices_by_type.cache_info()
```
davidverweij, line 49: I'd say this is the ACT block; perhaps do:
```python
result = inventory.all_devices_by_type.cache_info()
```
and then assert `result[0]` and `result[1]`

tests/data_transfer_tests/services/test_inventory.py
```diff
@@ -36,3 +36,34 @@ def test_normalise_day() -> None:
     result = one == two
 
     assert result
+
+
+def test_all_devices_by_type_cache_success() -> None:
+    inventory.all_devices_by_type.cache_clear()
+    num_requests = 10
+
+    with patch("requests.get", return_value=Mock()) as _:  # act
```
davidverweij, line 45: RE: `how come you have used a fixture to mock out requests? I discovered it's possible to do something similar but to override "requests.get" directly`

In my case I need a specific return value if you call something specific. Could indeed specify the returned value but using a mocked session with adapters allows for dynamic responses (which I used for the ByteFlies - but might not be as needed here)

tests/data_transfer_tests/services/test_inventory.py
```diff
@@ -36,3 +36,34 @@ def test_normalise_day() -> None:
...
+    inventory.all_devices_by_type.cache_clear()
+    num_requests = 10
+    response = MagicMock()
+    response.json = lambda: mock_inventory_devices_bytype
+
+    with patch("requests.get", return_value=response) as _:  # act
+        for __ in range(0, num_requests):
+            inventory.all_devices_by_type(utils.DeviceType.BTF)
+
+        hits, misses, _, __ = inventory.all_devices_by_type.cache_info()
```
davidverweij, line 67: Same here, see above.

davidverweij (commented):
LGTM - nice addition of tests and overall the performance is incredible! :rocket:.

Just a few minor comments, including one related to performance :wink:.

davidverweij:
❗ @jawrainey , I forgot to run tests; but it seems you haven't included the `mock_devices_bytype.json` in your commits.

tests/data_transfer_tests/services/test_inventory.py
jawrainey: You can set/obtain/change the response using [MagicMock](https://github.com/ideafast/middleware-services/blob/2e300907814126a968f1cfe344940a4924a7c541/tests/data_transfer_tests/services/test_inventory.py#L60-L61). So if I wanted, I could do `result = inventory.all_devices_by_type(utils.DeviceType.BTF)` and that would contain the result of `requests.get`. That way, you do not need sessions/adapters 🚀 

The reason I had to override `.json` (as a function using lambda) is because it gets called in `all_devices_by_type`

data_transfer/services/inventory.py
```diff
@@ -9,27 +9,40 @@
...
-    if not _response["meta"]["success"]:
-        return None
-    return _response["data"]["device_id"]
+def all_devices_by_type(device_type: utils.DeviceType) -> Any:
+    """
+    Retrieve complete list of ALL Devices by model.
+    This is cached as response can be quite large and will be used multiple times per DAG."""
+    model_id = dict(BTF=6, DRM=8)[device_type.name]
+    response = requests.get(f"{config.inventory_api}devices/bytype/{model_id}")
+    return response.json()
```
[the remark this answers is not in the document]
jawrainey: Could do, although at the moment this returns a list of Device objects rather than a list of Serial IDs so we can flexibly use this endpoint as needed.

Shall I change it such that we return only a set of Serial IDs instead?

data_transfer/services/inventory.py
```diff
@@ -9,27 +9,40 @@
 
 
 @lru_cache
```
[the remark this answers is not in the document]
jawrainey: Yeah, once the python script is finished running the memory will be cleared 🚀

data_transfer/services/inventory.py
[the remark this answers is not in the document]
davidverweij: Good point - fair enough! I'd then leave it as is.

data_transfer/devices/byteflies.py
```diff
@@ -260,8 +263,6 @@ def __patient_id_from_ucam(
         Determine PatientID by wear period of device in UCAM.
         """
         record = ucam.record_by_wear_period(device_id, start, end)
-        # TODO: inventory has small rate limit.
-        time.sleep(4)
```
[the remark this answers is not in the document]
jawrainey: For reference, this was likely a copy/paste issue from dreem as a call to the inventory is made in there 👍🏼

data_transfer/services/inventory.py
```diff
@@ -9,27 +9,40 @@
...
-    # TODO: validation
-    _response = response.json()
-    if not _response["meta"]["success"]:
-        return None
-    return _response["data"]["device_id"]
+def all_devices_by_type(device_type: utils.DeviceType) -> Any:
+    """
+    Retrieve complete list of ALL Devices by model.
+    This is cached as response can be quite large and will be used multiple times per DAG."""
+    model_id = dict(BTF=6, DRM=8)[device_type.name]
```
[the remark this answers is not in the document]
jawrainey: @davidverweij -- we may need to update for COS but for now those are the two devices who use inventory.

jawrainey:
@davidverweij -- i've addressed your comments, addeded the JSON mock responses, refactored inventory tests to use patch (so they're all using similar logic) and added tests for `services.inventory.device_history` as it was the only method not being tested. 

These changes only impact tests, so to verify: `poetry run nox -rs tests`

davidverweij (approved):
Nice 🚀  - thanks for accommodating the requested changes, and great work on the tests! Will be good to have a similar look at the test at other branches (e.g. BTF) at some point 👍

PR #64, opened 2021-04-01
restrict Record alteration if not in an appropriate stage, resolves #62

### LINKED ISSUE #62: Ensure _pipeline stages_ are ticked off in order
**Conversation from MS Teams**
@jawrainey: I found a bug in the current pipeline that will need addressed before going live as some data is marked as "is_downloaded=False" but is uploaded anyway 

@davidverweij: how about adding [validators](https://pydantic-docs.helpmanual.io/usage/validators/) to the [pipeline stages](https://github.com/ideafast/middleware-services/blob/06f58789b6b428d1209aeaae4a0ea89a206d4708/data_transfer/schemas/record.py#L34-L38) which ensures any next step cannot be true unless the previous one is. This prevents error in our coding where we - currently- could easily set any of these to true and mess around!
    
@jawrainey : Not sure how that maps to use within the pipeline, but can imagine having such validators being useful on the Record object -- would clean up similar logic in current db./main.py I guess

@davidverweij : Aye - this is at the record object (see link). Can throw (value) error when something/one is trying to set one true if the previous ones are not.

@jawrainey : Sounds good -- am not sure how it maps to the Record level as it depends on when they're (Record collections) are used in db/main.py I guess?  e.g., within /dags/btf.py, etc -- I guess being explicit and saying "Record.is_ready_for_upload" but might be trickier when working with lists of records? Either way, having solid validation is needed (laugh)

@davidverweij : It should affect all use of the Record class everywhere if the validator is in place at the Record scheme. I.e.; if some code tries to do record.is_uploaded = True and it doesn't pass validation; it fill error.

davidverweij:
These changes will prevent (in fact throw `ValueError`) if you try to progress a `pipeline_state` in a `Record` if the previous state has not been completed. Resolves #62 

# Testing
This won't affect any of the code as long as it doesn't violate the order of stages. However - this does mean we could consider `try / excepting` complete DAGs or perhaps sections within a DAG to actually catch if this happens possibly. Alternatively, this is more a safety-net for ourselves when coding.
Run:
```python
poetry run nox -r
```

### DIFF
--- a/data_transfer/schemas/record.py
+++ b/data_transfer/schemas/record.py
@@ -64,21 +64,23 @@ class Record(BaseModel):
 
     @validator("is_processed")
     def after_downloaded(cls, v: bool, values: dict) -> bool:
-        # values: a dict containing the name-to-value mapping of any previously-validated fields
-        # NOTE: only checks if the value 'v' is set to True
-        if v and not values["is_downloaded"]:
-            raise ValueError("NOT ALLOWED: this Record is not downloaded yet.")
-        return v
+        return cls.is_set_true_after(v, values, "is_downloaded")
 
     @validator("is_prepared", always=True)
     def after_processed(cls, v: bool, values: dict) -> bool:
-        if v and not values["is_processed"]:
-            raise ValueError("NOT ALLOWED: this Record is not processed yet.")
-        return v
+        return cls.is_set_true_after(v, values, "is_processed")
 
     @validator("is_uploaded")
     def after_prepared(cls, v: bool, values: dict) -> bool:
-        # NOTE: this value is normally set after upload; somewhat 'too little to late'
-        if v and not values["is_prepared"]:
-            raise ValueError("NOT ALLOWED: this Record is not prepared yet.")
-        return v
+        # NOTE: 'is_uploaded' is normally set after upload; somewhat 'too little to late'
+        return cls.is_set_true_after(v, values, "is_prepared")
+
+    @staticmethod
+    def is_set_true_after(set_value: bool, values: dict, required_true: str) -> bool:
+        # values: a dict containing the name-to-value mapping of any previously-validated fields
+        # only validate if trying to set to true
+        # then, previous step should also be true
+        if set_value and not values[required_true]:
+            step = required_true.split("_")[1]
+            raise ValueError(f"NOT ALLOWED: this Record is not {step} yet.")
+        return set_value
--- a/data_transfer/schemas/record.py
+++ b/data_transfer/schemas/record.py
@@ -32,10 +32,10 @@ class Record(BaseModel):
     dmp_folder: Optional[str]
 
     # Each stage of the pipeline
-    is_downloaded: Optional[bool] = False
-    is_processed: Optional[bool] = False
-    is_prepared: Optional[bool] = False
-    is_uploaded: Optional[bool] = False
+    is_downloaded: bool = False
+    is_processed: bool = False
+    is_prepared: bool = False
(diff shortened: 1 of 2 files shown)


data_transfer/schemas/record.py
```diff
@@ -32,10 +32,10 @@ class Record(BaseModel):
...
 
     # Each stage of the pipeline
-    is_downloaded: Optional[bool] = False
-    is_processed: Optional[bool] = False
-    is_prepared: Optional[bool] = False
-    is_uploaded: Optional[bool] = False
+    is_downloaded: bool = False
+    is_processed: bool = False
+    is_prepared: bool = False
+    is_uploaded: bool = False
```
davidverweij, line 38: I didn't understand why these were optional (i.e. could be assigned `None`) if we default them to `False` anyway..

data_transfer/schemas/record.py
```diff
@@ -57,3 +57,28 @@ def download_folder(self) -> Path:
...
+        if v and not values["is_processed"]:
+            raise ValueError("NOT ALLOWED: this Record is not processed yet.")
+        return v
+
+    @validator("is_uploaded")
+    def after_prepared(cls, v: bool, values: dict) -> bool:
+        # NOTE: this value is normally set after upload; somewhat 'too little to late'
+        if v and not values["is_prepared"]:
+            raise ValueError("NOT ALLOWED: this Record is not prepared yet.")
+        return v
```
jawrainey, line 84: Given the logic is the same across these three validators, I would pull it out into a public static method of the Record class, which should be accessible via `cls`.

```python
   @staticmethod
    def value_exists_or_raise(value_exists: bool, record: dict, value: str) -> bool:
        if value_exists and not record[value]:
            _type = value.split("-")[0]
            raise ValueError(f"NOT ALLOWED: this Record is not {_type} yet.")
        return value_exists
```

Then each validator method can call it (I assume) via:

```python
return cls.value_exists_or_raise(v, values, "is_downloaded")
```

This also means you can test the if statement -- so two extra tests: one for fail and one for success.

tests/data_transfer_tests/schemas/test_record.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.mark.xfail(raises=ValueError)
+def test_not_allowed_uploaded(get_record: Record) -> None:
+    get_record.is_downloaded = True
+    get_record.is_processed = True
+
+    get_record.is_uploaded = True  # act
+
+
+@pytest.mark.xfail(raises=ValueError)
+def test_not_allowed_processed(get_record: Record) -> None:
```
jawrainey, line 51: I'd also test some ways that it passes, perhaps by testing `value_exists_or_raise` above?

jawrainey (changes requested):
> This won't affect any of the code as long as it doesn't violate the order of stages. However - this does mean we could consider try / excepting complete DAGs or perhaps sections within a DAG to actually catch if this happens possibly. Alternatively, this is more a safety-net for ourselves when coding.

I would not add try/except around DAGS but instead let it fail. These changes are primarily for developers rather than solving known bugs, i.e., they do not resolve the issue I note in #62 

I've tested this locally by running the tests and also running the DRM pipeline where the bug exists -- an error is thrown which is great and helped confirm where the BUG is. I'll create a seperate PR to resolve the bug.

davidverweij:
@jawrainey, I've pulled out the method to static as indicated. @ColinBD, can you review this PR overall; and merge if ok? Cheers!

ColinBD:
All good.

PR #66, opened 2021-04-06
Control BTF authentication via env. variables; cache part of BTF API; resolves #60

### LINKED ISSUE #60: Reauthorize BTF pipeline
### Problem

Recent dev deployment run of the BTF pipeline resulted in:

> 2021-03-31 01:02:22 [15] [data_transfer.lib.byteflies:ERROR] GET Exception to {URL}/recordings/{UUID} 
Traceback (most recent call last):
  File "/app/data_transfer/lib/byteflies.py", line 150, in __get_response
    response.raise_for_status()
  File "/usr/local/lib/python3.8/site-packages/requests/models.py", line 943, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: {URL}/recordings/{UUID} 

This is likely because BTF uses one object that shared authorisation and the pipeline may run for longer than the pre-defined 1-hour authroisation limit. Therefore we need to modify [the following](https://github.com/ideafast/middleware-services/blob/master/data_transfer/devices/byteflies.py#L63-L75) such that any time a request is made that tokens will be regenerated if needed. 

### Potential solution

We could modify each method (`download_file` and `download_metadata`) to call `authenticate` before it performs logic to ensure if a token has expired that a new one will be generated? I propose that we add a datetime to the constructor of the Byteflies class, then add a new method to do the auth to make it explict as below. We could of course put this logic in `authenticate` if prefered.

```python

def __reauthenticate(self):
    if self.token_created_timestamp <= 60_mins_ago:
        self.authenticate()
``` 

If we go with this appraoch, we can then test this by calling a method that requires authentication (e.g., `download_file`), reassign `self.token_created_timestamp` to 80 mins and verify that the token is different in `self.session`.

davidverweij:
In this PR, I've restructured BTF's authentication from being a class attribute, to being stored in the environmental variables; much like how this is done at the `dmpy` package. This will ensure that the authentication is re-done every 50 minutes (10 minutes before a token expires) - this check is triggered when preparing for an actual GET request. The proposed solution moved the _authentication_ to the `lib/byteflies.py` and thereby also removes the need to pass a `requests.Session` variable around. 

Additionally, I've cached the `__get_recording_by_id()` (maxsize of 1). Since we wanted the code to run for any individual file, at any time (no dependencies on living variables), the pipeline made subsequent calls to the exact same API endpoint. With the proposed changes, the result of this call is cached for max 5 minutes (since the links in the payload are valid for only 10). A quick test shows (for example) out of 21 files to be downloaded from one recording, 7 files to be downloaded were based on the same API call and have now been reduced to one. The other 14 calls could not be cached as they call a deeper API link to retrieve the algorithm download url; which has to be done for each algorithm file. I've turned down the `sleep.time()` from 1 to .5 to speed this up a bit.

## Testing
- `poetry run nox -r`
- set env's to point to local mongoDB
- clear mongo,DB data and upload folder
- `poetry run consumer`
- `python data_transfer/main.py BTF Muenster 0 3`
- (increase the _3_ to keep adding records; i.e. go further back in time)

Any feedback highly appreciated @jawrainey @ColinBD !

### DIFF
--- a/data_transfer/devices/byteflies.py
+++ b/data_transfer/devices/byteflies.py
@@ -55,6 +55,22 @@ class Byteflies:
         self.study_site = study_site
         self.device_type = utils.DeviceType.BTF
         self.file_type = ".csv"
+        # timestamp used to optimise lru_cache in file download pipeline
+        self.last_downloaded = 0
+        self.download_cache_expiry = 300  # 5 minutes
+
+    def __get_timestamp(self) -> int:
+        """
+        return a timestamp that changes every x minutes to allow caching of API call
+        # NOTE: only used in the download stage
+        """
+        now = int(datetime.now().timestamp())
+        self.last_downloaded = (
+            now
+            if now - self.last_downloaded > self.download_cache_expiry
+            else self.last_downloaded
+        )
+        return self.last_downloaded
 
     def download_metadata(self, from_date: int, to_date: int) -> None:
         """
@@ -184,6 +200,7 @@ class Byteflies:
             record.meta["recording_id"],
             record.meta["signal_id"],
             record.meta["algorithm_id"],
+            self.__get_timestamp(),
         )
 
         # filename if succes, None if not
--- a/data_transfer/devices/byteflies.py
+++ b/data_transfer/devices/byteflies.py
@@ -4,8 +4,6 @@ from datetime import datetime
 from pathlib import Path
 from typing import Dict, List, Optional
 
-import requests
-
 from data_transfer import utils
 from data_transfer.config import config
 from data_transfer.db import all_hashes, create_record, read_record, update_record
@@ -55,23 +53,8 @@ class Byteflies:
         Authenticate with AWS cognito to access ByteFlies resources
         """
         self.study_site = study_site
-        self.session = self.authenticate()
         self.device_type = utils.DeviceType.BTF
         self.file_type = ".csv"
-
-    def authenticate(self) -> requests.Session:
-        """
-        Authenticate once when object created to share session between requests
-        """
-        credentials = dict(
-            username=config.byteflies_username,
-            password=config.byteflies_password,
-            client_id=config.byteflies_aws_client_id,
-        )
-        token = byteflies_api.get_token(credentials)
-        session = byteflies_api.get_session(token)
-        log.info("Authentication successful")
-        return session
 
     def download_metadata(self, from_date: int, to_date: int) -> None:
         """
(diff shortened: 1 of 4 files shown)


davidverweij:
**NOTE**, a few things are left to do; for which I have created Issue #67

data_transfer/devices/byteflies.py
```diff
@@ -55,23 +53,24 @@ def __init__(self, study_site: StudySite) -> None:
...
         self.study_site = study_site
-        self.session = self.authenticate()
         self.device_type = utils.DeviceType.BTF
         self.file_type = ".csv"
+        # timestamp used to optimise lru_cache in file download pipeline
+        self.last_downloaded = 0
+        self.download_cache_expiry = 300  # 5 minutes
 
-    def authenticate(self) -> requests.Session:
+    def __get_timestamp(self) -> int:
```
jawrainey, line 62: 🚀

data_transfer/lib/byteflies.py
```diff
@@ -13,46 +16,52 @@
...
-    session = requests.Session()
-    session.headers.update({"Authorization": f"{token}"})
-    return session
+            os.environ["UCAM_ACCESS_TOKEN"] = access_token
+            os.environ["UCAM_ACCESS_TOKEN_GEN_TIME"] = str(now)
+
+        except Exception:
+            log.error("Exception:", exc_info=True)
+
+    return os.getenv("UCAM_ACCESS_TOKEN")
```
jawrainey, line 61: I assume you meant "BTF_*"?

data_transfer/lib/byteflies.py
```diff
@@ -13,46 +16,52 @@
...
-                "USERNAME": f"{creds['username']}",
-                "PASSWORD": f"{creds['password']}",
-            },
-        },
-    )
-    #  TODO: catch/log exception
-    res.raise_for_status()
-    resp = res.json()
-    return str(resp["AuthenticationResult"]["IdToken"])
+def __btf_access_token(forced: bool = False) -> str:
```
jawrainey, line 19: I'd make it public (`btf_access_token`) for direct testing in the future 🚀

jawrainey (changes requested):
Worked for me locally with no issues. There's one minor issue with setting environmental variables, but otherwise perfect 👍🏼

data_transfer/lib/byteflies.py
davidverweij: 😭  yes..

PR #69, opened 2021-04-08
Use latest DMPy version

jawrainey:
The DMP returns a list of errors as a response if an error occurs. Therefore, if a file is attempted to be uploaded and an error occurs, then it is marked as uploaded because the prior dmpy logic did not account for this. This was resolved in [dmpy#7](https://github.com/ideafast/dmpy/pull/7) and so this PR adds the latest package with the changes.

### Testing

1. Review changes
2. Run `poetry install` and `poetry run nox` locally

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -180,7 +180,7 @@ python-versions = "*"
 
 [[package]]
 name = "dmpy"
-version = "0.1.0"
+version = "0.1.3"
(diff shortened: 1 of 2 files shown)


davidverweij (approved):
I had to double check why you removed the `pymongo` dev-dependency, but after testing everything I realised we also have it as a standard dependency... 🕵️  

so all good to me!

PR #76, opened 2021-04-13
only run GH tests on PR

davidverweij:
As we have a consistent flow of creating PRs for any changes, we'd discussed to only run Github Action Tests on a PR to reduce the amount of tests; especially since we push our changes regularly and knowing that tests will fail in these intermediate pushes.


jawrainey (approved):
🚀

PR #71, opened 2021-04-09
GitHub Action for Tests

jawrainey:
Automatically run tests via GitHub Actions when pushed to origin, on PR, or manually (if desired). This PR closes #39 and [saves us from having to run tests locally](https://github.com/ideafast/middleware-services/pull/63#pullrequestreview-628778191).

Changes:

- Set default values for consumer so tests can load settings
- Moved loading of dotenv in consumer to when settings are invoked
- Added default values for database on `data_transfer`

**Note:** I had to modify BTF tests because they depend on internals of the library. Imo, this should ideally be refactored in another PR.

### Testing

- Review code changes
- Review the Actions tab and see the workflow

PR #77, opened 2021-04-13
fix mypy typing error `config.settings`

davidverweij:
Interestingly, PR #71 introduced a typing error due to `GlobalConfig` (logically) not having all attributes from `Settings`. However, `settings()` could now return either type. After some experimentation with @jawrainey, it was clear that using `Union[Settings, GlobalConfig]` was not working.

Strangely, the `mypy` session in `poetry run nox -r` does not pick up on this error (and therefore neither does the Github Actions Tests), but the `mypy` session in `pre-commit` as well as `mypy data_transfer` does.

As Jay suggested, although not ideal, using `Any` would give us a quick fix.

### DIFF
--- a/data_transfer/config.py
+++ b/data_transfer/config.py
@@ -1,6 +1,7 @@
 import os
 from functools import lru_cache
 from pathlib import Path
+from typing import Any
 
 from dotenv import get_key, load_dotenv
 from pydantic import BaseSettings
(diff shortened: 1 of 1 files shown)


jawrainey (approved):
Perfect! Quite surprised that inheritance types in mypy didn't work as expected. Possibly we could configure mypy to account for this, but using Any seems a good solution given we don't depend on the type returned from settings anywhere.

PR #80, opened 2021-04-16
Inventory datetime fix

### LINKED ISSUE #79: Inventory history not correct when (faulty) duplicate checkouts occur
Occasionally (and unexplainably) the inventory allows multiple checkouts at the near-exact same time. Currently, the inventory [`device_history`](https://github.com/ideafast/middleware-services/blob/d84e93004ae6fd1326b58abe73f2c28a11fcb00f/consumer/routers/inventory.py#L68-L97) endpoint only expects a single checkout, and a single checkin. For example, an device history might look like:

```
2020-11-25 09:37:12 AM | checkin
2020-11-10 11:24:06 AM | checkout
2020-11-10 11:24:05 AM | checkout
```
The current implementation currently doesn't look at the 'type' of event (i.e. checkout or check in) and thereby falsly assumes (in the above case) that the device was only checked out for one second.

@jawrainey has a fix ready to I believe!

jawrainey:
As @davidverweij outlined in #79 it is possible for a user of snipe-it to button-smash and make N checkout/checkin requests at the same time. This then causes issues with existing logic, resulting in incorrect datetimes in the history of devices.

As such, I have modified the code to fix this bug. While the fix is potentially simple (one-line), I instead refactored the method because it was performing too much logic implictely, which while more efficient impacted readability. Likewise, I have taken this as a chance to pull out `serialize` methods to relevant schema classes.

### To Test

1. Review Code
2. Review Tests from C.I
3. Download locally and run `poetry run consumer` and navigate to to the device that caused the bug [http://0.0.0.0:8000/inventory/device/history/BTF-ABC](http://0.0.0.0:8000/inventory/device/history/BTF-ABC]). The history should now be resolved.

Closes #79


consumer/schemas/inventory.py
```diff
@@ -0,0 +1,52 @@
...
+    id: int
+    serial: str
+    device_id: str
+    is_checkout: bool
+    model: Optional[str]
+    manufacturer: Optional[str]
+    location: Optional[str]
+
+    @classmethod
+    def serialize(cls, device: dict) -> "Device":
```
davidverweij, line 16: Rather than using a forward reference, I argue you can use 
```python
from __future__ import annotations
```
at the very top of this file. It will then automatically detect future reference and you don't need to provide them as a string anymore. The benefit is that future refs will be integrated in python soon, and then we only need to remove this import statement (though I expect it doesn't do any harm and thus upgrading will go smoothly). See also [here](https://stackoverflow.com/questions/61544854/from-future-import-annotations), I've just tried this and it works nicely.

consumer/schemas/inventory.py
```diff
@@ -0,0 +1,52 @@
...
+            location=name_or_none(device["location"]),
+        )
+
+
+class HistoryItem(BaseModel):
+    patient_id: str
+    datetime: str
+
+    @classmethod
+    def serialize(cls, device: dict) -> "HistoryItem":
```
davidverweij, line 38: See comment above

consumer/schemas/inventory.py
```diff
@@ -0,0 +1,52 @@
...
+    is_checkout: bool
+    model: Optional[str]
+    manufacturer: Optional[str]
+    location: Optional[str]
+
+    @classmethod
+    def serialize(cls, device: dict) -> "Device":
+        """Simplifies reuse across inventory API."""
+
+        def name_or_none(item: dict) -> Optional[str]:
```
davidverweij, line 19: This method feels a bit redundant imho. `dict.get("name", None)` already returns None if not found. Even without `None`, i.e.g `dict.get("name")` already returns `None` by default.

For readability, I'd argue to replace the use of `name_or_none()` below with just the `dict.get("name")` method; unless I am missing something that gives an advantage here! 

i.e.: 
```python
model=device["model"].get("name"),
manufacturer=device["manufacturer"].get("name"),
location=device["location"]).get("name"),
```

consumer/schemas/inventory.py
```diff
@@ -0,0 +1,52 @@
...
+    patient_id: str
+    datetime: str
+
+    @classmethod
+    def serialize(cls, device: dict) -> "HistoryItem":
+        """Device is an item from a snipe-it response.
+        This also contains an 'action_type' that is either 'checkout'
+        or 'checkin from' but is not currently used."""
+        return cls(
+            patient_id=device["target"]["name"].strip(),
```
davidverweij, line 43: You could use the `dict.get("name")` here too if applicable.

consumer/routers/inventory.py
```diff
@@ -1,39 +1,22 @@
...
-    )
-
-
 @router.get("/devices/bytype/{model_id}")
 async def devices_by_type(model_id: int) -> List[Device]:
     """Retrieve metadata about ALL devices for by a specific model."""
     url = f"hardware?limit=500&model_id={model_id}"
     res = await inventory.response(url)
-    return [serialize_device(i) for i in res["rows"]]
+    return [Device.serialize(i) for i in res["rows"]]
```
davidverweij, line 19: I'm liking the move to a static method in the class :rocket:!

tests/consumer_tests/routers/test_inventory.py
```diff
@@ -85,6 +86,25 @@ async def mock_get(path: str, params: str = None):
     assert result["T-456"]["checkout"] and result["T-456"]["checkin"]
 
 
+def test_device_history_with_device_in_use_fail(
```
davidverweij, line 89: This test is called 'in_use_fail', but can't see how it should in the asserts? Is this expected to fail (`xfail`)? Because it doesn't in the test. Perhaps rewording the name will help.

tests/consumer_tests/routers/test_inventory.py
```diff
@@ -94,7 +114,7 @@ def test_serialize_device_required_params(key, response_row) -> None:
     del response_row[key]
 
     with pytest.raises(KeyError):
```
davidverweij, line 116: Could consider using `@pytest.mark.xfail(raises=KeyError, strict=True)`, but up to you!

davidverweij (commented):
LGTM :rocket:! Tested it and works well. One comment regarding future references that is worth taking a look at.

consumer/schemas/inventory.py
jawrainey: > For readability, I'd argue to replace the use of name_or_none() below with just the dict.get("name") method; unless I am missing something that gives an advantage here!

I agree, however, the issue was that the dictionary parameter may be `None` rather than `{}`. So `device["model"].get("name")` would throw an `Attribute Error`. [This was added in this commit here](https://github.com/ideafast/middleware-services/pull/13/commits/057f4dbc12088531e003b94ba071351fda3cd28f). You also cannot chain `.get` calls, e.g., `device.get("model").get("name")` although we could if the default is a dict `device.get("model", {}).get("name")`

Shall I update the type of the argument to optional and add a comment to explain this need?

tests/consumer_tests/routers/test_inventory.py
jawrainey: I think I had planned to test it one way but went for the successful route instead, so will rename to `test_device_history_with_device_multiple_checkouts`

consumer/schemas/inventory.py
jawrainey: I agree, but in this case we probs don't want a default name (patient_id) so letting it throw if it doesn't exist may be ideal?

consumer/schemas/inventory.py
davidverweij: Ah I see, that makes sense. Happy to continue with `name_or_none` if you can add a quick comment for its use case 👍

jawrainey: Went with the chaining option 🚀

davidverweij:
Thanks for the changes @jawrainey! 🚀

PR #86, opened 2021-04-23
Use None when serializing inventory response

### LINKED ISSUE #85: Assign correct default values to inventory
## Problem

As described in #80, there was once a method named `name_or_none` in the inventory to ensure dictionary items are correctly serialised. This method raised the following discussion:

> This method feels a bit redundant imho. `dict.get("name", None)` already returns None if not found. Even without `None`, i.e.g `dict.get("name")` already returns `None` by default.
For readability, I'd argue to replace the use of `name_or_none()` below with just the `dict.get("name")` method; unless I am missing something that gives an advantage here! 
>i.e.: 
```python
model=device["model"].get("name"),
manufacturer=device["manufacturer"].get("name"),
location=device["location"]).get("name"),
```

_Originally posted by @davidverweij in https://github.com/ideafast/middleware-services/pull/80#discussion_r615717060_

The solution implemented was to go with a chaining option `location=device.get("location", {}).get("name") if device else None,` which is incorrect as `location` _always_ exist in the response, but may be `None` if it is not set in the inventory. This breaks the current inventory use.

## Solution

1. Re-add `name_or_none` method

jawrainey:
Detailed outlined in #85, but in brief we're trying to access attributes of a dictionary that doesn't exist (it's `None`).

## To reproduce error

1. `git checkout master`
2. `poetry run consumer`
2. Visit [http://0.0.0.0:8000/inventory/devices/bytype/8](http://0.0.0.0:8000/inventory/devices/bytype/8) and observe `Internal Server Error`

## Solution

To see the change checkout this branch, run tests (I've added a new test to confirm this bug), and navigate to the above URL. All should be fine. Closes #85

### DIFF
--- a/consumer/schemas/inventory.py
+++ b/consumer/schemas/inventory.py
@@ -17,14 +17,19 @@ class Device(BaseModel):
     @classmethod
     def serialize(cls, device: dict) -> Device:
         """Simplifies reuse across inventory API."""
+
+        def name_or_none(item: dict) -> Optional[str]:
+            """Required as sometimes item may be None as it is a dictionary value."""
+            return item.get("name", None) if item else None
+
         return cls(
             id=device["id"],
             serial=device["serial"].replace(" ", ""),
             device_id=device["asset_tag"],
             is_checkout=device["status_label"]["status_meta"] == "deployed",
-            model=device.get("model", {}).get("name"),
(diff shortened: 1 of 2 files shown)


davidverweij (approved):
LGTM - see thoughts on GH Issue

PR #92, opened 2021-05-06
Camcog

ColinBD:
We now pop GDPR concerning data from the dictionary prior to upload.


jawrainey:
Tested locally from master and this branch and can confirm the changes work

PR #97, opened 2021-07-01
fix unknown IDs

davidverweij:
As @ColinBD indicated, the location of the IDEAFAST valid, or alias, ID can be in a different location in the list of data entries. Originally, we checked for the first two entries - though it seemed, strangely only occurring for one study site, that TFA data can contain a third item in the list, which for a set of patients contained their ID's. I've rewritten the snippet that checks for the ID which now iterates to as many as needed until it finds a valid ID or known alias.

### DIFF
--- a/data_transfer/lib/thinkfast.py
+++ b/data_transfer/lib/thinkfast.py
@@ -73,18 +73,16 @@ def id_in_whitelist(input_ID: str) -> Optional[str]:
     return output_ID
 
 
-def get_participant_id(subjectItems: Dict[int, Any]) -> Optional[str]:
-    # find the ideaFast id in the json...
-    # this is a bit tricky as it's not always in the same place.
+def get_participant_id(subjectItems: dict) -> Optional[str]:
+    # the ID can be placed in any of the list items. Iterate untill found.
     ideaId = ""
 
-    # The line below is a horribly shakey solution!
-    if len(subjectItems[2]["text"]) == 7 and (" " not in subjectItems[2]["text"]):
-        ideaId = subjectItems[2]["text"]
-    else:
-        ideaId = subjectItems[1]["text"]
-    # validate ID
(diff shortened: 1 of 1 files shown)


davidverweij:
To get the missing files to WP4 quickly, I've build and published [a docker image](https://hub.docker.com/layers/ideafast/middleware-dtransfer/0.1.10-DEV/images/sha256-bc65e1b7d404868c98b4253fd1580284414a1da497f03fa950c19a5ced4ef5ca?context=repo) on this branch and set our `dtransfer_live` to run on that in the mean-time. When merged, it might be good to update the docker image or at least keep this in mind.

data_transfer/lib/thinkfast.py
```diff
@@ -73,18 +73,16 @@ def id_in_whitelist(input_ID: str) -> Optional[str]:
     return output_ID
 
 
-def get_participant_id(subjectItems: Dict[int, Any]) -> Optional[str]:
-    # find the ideaFast id in the json...
-    # this is a bit tricky as it's not always in the same place.
+def get_participant_id(subjectItems: dict) -> Optional[str]:
+    # the ID can be placed in any of the list items. Iterate untill found.
     ideaId = ""
```
jawrainey, line 78: Might be useful to make this None rather than "" so it's explict that if no ID is found that None exists? This does not impact the use of `get_participant_id` below since `if ""` and `if None` both result in same logic.

jawrainey (commented):
Great work @davidverweij. I've verified this through looking on the DMP and now Rotterdam data exists. I made one minor (unrelated) comment regarding how ideaId is defined.

data_transfer/lib/thinkfast.py
davidverweij: Agreed 👍  - see commit below

